PR #43, opened 2021-05-03
added per block rewards minting

catalinnnn:


elrond_dex_farm/src/lib.rs
```diff
@@ -536,14 +543,16 @@ pub trait Farm {
         }
     }
 
-    #[view(simulateEnterFarm)]
+    #[endpoint(simulateEnterFarm)]
```
claudiulataretu, line 546: simulateEnterFarm should only calculate the minted rewards and not actually mint them

elrond_dex_farm/src/lib.rs
```diff
@@ -561,7 +570,7 @@ pub trait Farm {
         })
     }
 
-    #[view(simulateExitFarm)]
+    #[endpoint(simulateExitFarm)]
```
claudiulataretu, line 573: simulateExitFarm should only calculate the minted rewards and not actually mint them

claudiulataretu (changes requested):

claudiulataretu (approved):

PR #49, opened 2021-05-08
Fix initial liquidity locking

catalinnnn:
This is the correct way we want to lock the initial liquidity. Before this commit, because all the information is stored inside the SFT, the first "farmer" could exit farm and actually get all his initial tokens. This is the normal case except for the first one. For the first one, because of liquidity calculations, we need to permanently lock a part of first input tokens. 
In this fix, the liquidity is locked only after the SFT attributes are crafted, which is essential. Now the first minimum amount will never be moved.

### DIFF
--- a/elrond_dex_farm/src/lib.rs
+++ b/elrond_dex_farm/src/lib.rs
@@ -221,10 +221,10 @@ pub trait Farm {
             entering_epoch: self.blockchain().get_block_epoch(),
         };
 
-        // Do the actual permanent lock of first minimul liquidity
+        // Do the actual permanent lock of first minimum liquidity
         // only after the token attributes are crafted for the user.
         if is_first_provider {
-            liquidity -= BigUint::from(self.liquidity_pool().minimul_liquidity_farm_amount());
+            liquidity -= BigUint::from(self.liquidity_pool().minimum_liquidity_farm_amount());
         }
 
         // This 1 is necessary to get_esdt_token_data needed for calculateRewardsForGivenPosition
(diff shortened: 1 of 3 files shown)


elrond_dex_farm/src/lib.rs
```diff
@@ -220,6 +221,12 @@ pub trait Farm {
             entering_epoch: self.blockchain().get_block_epoch(),
         };
 
+        // Do the actual permanent lock of first minimul liquidity
+        // only after the token attributes are crafted for the user.
+        if is_first_provider {
+            liquidity -= BigUint::from(self.liquidity_pool().minimul_liquidity_farm_amount());
```
claudiulataretu, line 227: I think you have a typo here: `minimul_liquidity_farm_amount`
Shouldn't it be `minimum_liquidity_farm_amount` ?

elrond_dex_farm/src/lib.rs
```diff
@@ -649,13 +654,19 @@ pub trait Farm {
...
+        let mut liquidity = self.liquidity_pool().calculate_liquidity(
             &farm_contribution,
             &farming_pool_token_id,
             &token_in,
         );
         let farm_token_id = self.farm_token_id().get();
         let farming_pool_token_nonce = self.farm_token_nonce().get();
+
+        if is_first_provider {
+            liquidity -= BigUint::from(self.liquidity_pool().minimul_liquidity_farm_amount());
```
claudiulataretu, line 667: Same typo as above

elrond_dex_farm/src/liquidity_pool.rs
```diff
@@ -92,14 +94,21 @@ pub trait LiquidityPoolModule {
...
             let total_reserves = virtual_reserves + actual_reserves + reward_amount;
             amount * &total_supply / total_reserves
         }
     }
 
+    fn is_first_provider(&self) -> bool {
+        self.total_supply().get() == 0
+    }
+
+    fn minimul_liquidity_farm_amount(&self) -> u64 {
```
claudiulataretu, line 108: Same typo as above

claudiulataretu (changes requested):

claudiulataretu (approved):

PR #52, opened 2021-05-11
Audit fixing

catalinnnn:
Fixed a miscalculation of special fee. The total fee was computed correctly but was not split correctly between providers and special fee.

### DIFF
--- a/elrond_dex_pair/src/amm.rs
+++ b/elrond_dex_pair/src/amm.rs
@@ -58,18 +58,8 @@ pub trait AmmModule {
         (numerator / denominator) + BigUint::from(1u64)
     }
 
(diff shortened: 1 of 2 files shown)


elrond_dex_pair/src/amm.rs
```diff
@@ -58,20 +58,10 @@ pub trait AmmModule {
         (numerator / denominator) + BigUint::from(1u64)
     }
 
-    fn get_special_fee_from_fixed_input(&self, amount_in: BigUint) -> BigUint {
+    fn get_special_fee_fron_input(&self, amount_in: BigUint) -> BigUint {
```
claudiulataretu, line 61: typo: get_special_fee_fron_input -> get_special_fee_from_input

PR #59, opened 2021-05-17
New farm refactor

catalinnnn:


elrond_dex_farm/src/lib.rs
```diff
@@ -175,6 +182,10 @@ pub trait Farm:
             "Cannot exit farm with 0 entering amount"
         );
 
+        // Before removing liquidity, first generate the rewards.
+        let reward_token_id = self.reward_token_id().get();
```
sasurobert, line 186: so here was a logical issue, not ?

sasurobert: Can you add a unit test ?

sasurobert (approved):

elrond_dex_farm/src/lib.rs
catalinnnn: Not a logical issue. There is a test for this, but tests reside inside arwen repo, at least atm

PR #60, opened 2021-05-17
Double APR for lock rewards option

catalinnnn:


elrond_dex_farm/src/lib.rs
```diff
@@ -9,6 +9,7 @@ type Nonce = u64;
 const PENALTY_PERCENT: u64 = 10;
 const BURN_TOKENS_GAS_LIMIT: u64 = 5000000;
 const EXIT_FARM_NO_PENALTY_MIN_EPOCHS: u64 = 3;
+const LOCKED_REWARDS_LIQUIDITY_MUTIPLIER: u64 = 2;
```
sasurobert, line 12: we need functions through which these values can be set. Also this value has to be written in the created NFT for the farming position.

elrond_dex_farm/src/lib.rs
```diff
@@ -9,6 +9,7 @@ type Nonce = u64;
 const PENALTY_PERCENT: u64 = 10;
```
sasurobert, line 9: set function for PenalyPercent is need as well.

elrond_dex_farm/src/lib.rs
```diff
@@ -118,8 +152,20 @@ pub trait Farm:
...
 
+        // For locked rewards option, exit before minumum epochs is not allowed
+        require!(
+            !farm_attributes.with_locked_rewards
+                || !self.should_apply_penalty(farm_attributes.entering_epoch),
+            "Exit too early for lock rewards option"
+        );
+        require!(
+            !farm_attributes.with_locked_rewards
+                || enter_amount >= Self::BigUint::from(LOCKED_REWARDS_LIQUIDITY_MUTIPLIER),
```
sasurobert, line 163: why are you comparing with the value "2" instead of value 0 ?

elrond_dex_farm/src/lib.rs
```diff
@@ -134,26 +180,43 @@ pub trait Farm:
...
+            self.send()
+                .burn_tokens(&farming_token_id, 0, &penalty_amount, BURN_TOKENS_GAS_LIMIT);
             enter_amount -= penalty_amount;
         }
 
-        self.send()
-            .transfer_tokens(&farming_token_id, 0, &enter_amount, &caller);
-        self.send()
-            .transfer_tokens(&reward_token_id, 0, &reward, &caller);
+        let mut reward_nonce = 0u64;
```
sasurobert, line 193: take out line 193-210 into a separate function

elrond_dex_farm/src/lib.rs
```diff
@@ -202,30 +265,59 @@ pub trait Farm:
...
             self.create_farm_tokens(&re_added_liquidity, &farm_token_id, &new_attributes);
         self.send()
             .transfer_tokens(&farm_token_id, new_nonce, &re_added_liquidity, &caller);
 
         // Send rewards
-        self.send()
-            .transfer_tokens(&reward_token_id, 0, &reward, &caller);
+        let mut reward_nonce = 0u64;
+        if farm_attributes.with_locked_rewards {
+            if reward > 0 {
```
sasurobert, line 278: add a require instead of the IF.

sasurobert (changes requested):

elrond_dex_farm/src/lib.rs
catalinnnn: because of the division happening a few line below. if the initial value is 1, then after the division with 2 it will still be 0

elrond_dex_farm/src/lib.rs
catalinnnn: a require would end the execution and fail the transaction whereas in case of 0 rewards, we still need to send the LP tokens (for example) back

elrond_dex_farm/src/lib.rs
catalinnnn: done

elrond_dex_farm/src/lib.rs
catalinnnn: refactored to be more intuitive

elrond_dex_farm/src/lib.rs
catalinnnn: refactored to be more intuitive

elrond_dex_farm/src/lib.rs
catalinnnn: done

elrond_dex_farm/src/lib.rs
catalinnnn: done

sasurobert (approved):

PR #62, opened 2021-05-22
Farm rewards

catalinnnn:
New method of calculating per block rewards in farming contract.
The key idea is to have a reward_per_share variable that can only increase and each time someone enters farm, we keep that variable inside the SFT attributes. When exiting, we get the new reward_per_share, and the reward amount will be amount * (reward_per_share - attributes.reward_per_share)
The nice thing about this method is that each farmer gets exactly the amount of reward generated in his farming period (with respect to his position in farm, of course)
The downside of this method is that the variable keeps increasing to infinity and we use that 1e12 value to avoid having subunitary results of division.
Tested in mandos.


dex/elrond_dex_farm/src/lib.rs
```diff
@@ -161,43 +162,43 @@ pub trait Farm:
...
+        self.generate_aggregated_rewards(&reward_token_id);
 
-        let caller = self.blockchain().get_caller();
-        let mut reward = self.remove_liquidity(&liquidity, &enter_amount)?;
-        let farming_token_id = self.farming_token_id().get();
-        self.send()
-            .burn_tokens(&farm_token_id, token_nonce, &liquidity, burn_gas_limit);
+        let mut reward = self.calculate_reward(
+            &amount,
+            &self.reward_per_share().get(),
```
sasurobert, line 176: is reward_per_shard derived from rewards_per_block and total_liquidity ?

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -368,35 +363,33 @@ pub trait Farm:
...
-        require!(farm_token_supply > liquidity, "Not enough supply");
+        require!(farm_token_supply >= amount, "Not enough supply");
 
-        let attributes = self.decode_attributes(&attributes_raw)?;
-        require!(
-            liquidity <= attributes.total_liquidity_amount,
-            "Bad arguments"
-        );
+        let current_block = self.blockchain().get_block_nonce();
+        let to_be_minted = self.calculate_per_block_rewards(current_block);
```
sasurobert, line 382: in case of farm we need a vector of nonce / rewards change as we might change the rewards per block multiple times.

dex/elrond_dex_farm/src/rewards.rs
```diff
@@ -1,41 +1,37 @@
 elrond_wasm::imports!();
 elrond_wasm::derive_imports!();
+use super::config;
 
 type Nonce = u64;
+type Epoch = u64;
+const DIVISION_SAFETY_CONSTANT: u64 = 1000000000000;
```
sasurobert, line 7: i think it would be better to have it in storage - and do storageLoadAsBigInt.

dex/elrond_dex_farm/src/rewards.rs
```diff
@@ -44,36 +40,113 @@ pub trait RewardsModule {
...
-            big_zero.clone()
-        };
+        if total_reward > 0 {
+            self.increase_reward_reserve(&total_reward);
+            self.update_reward_per_share(&total_reward);
+        }
+    }
+
+    fn reset_temporary_fee_storage(&self) -> Self::BigUint {
+        let current_block = self.blockchain().get_block_nonce();
```
sasurobert, line 55: is this still needed ? these are fees from penalties ?

sasurobert (changes requested):

dex/elrond_dex_farm/src/lib.rs
sasurobert: from excel examples this is not needed. as the rewards are actually given from rewards reserve. On update reward rate we do generate rewards and update price per share.

dex/elrond_dex_farm/src/lib.rs
catalinnnn: reward per share is an accumulator. at each state it adds the rewards that were generated divided by current total liquidity

dex/elrond_dex_farm/src/rewards.rs
catalinnnn: these are the fees from swap pairs. those should be distributed with the same logic if we decide to send fees from swap pairs to farm contracts

dex/elrond_dex_farm/src/lib.rs
catalinnnn: done

dex/elrond_dex_farm/src/rewards.rs
catalinnnn: done

sasurobert (approved):

sasurobert (approved):

andrei-marinica (approved):

PR #65, opened 2021-05-25
made dex proxy non payable

catalinnnn:


claudiulataretu (approved):

dex/elrond_dex_pair/src/lib.rs
```diff
@@ -433,32 +477,59 @@ pub trait Pair:
...
+                    accept_funds_func.as_slice(),
+                    self.transfer_exec_gas_limit().get(),
+                ),
+                OptionalArg::None => {
+                    let no_func: &[u8] = &[];
+                    (no_func, 0u64)
+                }
+            };
+
+            let result = self.send().direct_esdt_execute(
```
sasurobert, line 516: direct_esdt_execute is not good - this works intra shard only. You should use transfer and execute and should not care about the results - as those may be different from intra and cross shard. In case of cross shard there are no results.

dex/elrond_dex_router/src/factory.rs
```diff
@@ -36,7 +36,7 @@ pub trait FactoryModule {
     ) -> SCResult<Address> {
         require!(self.pair_code_ready().get(), "Pair code not ready");
         let code_metadata = CodeMetadata::UPGRADEABLE;
-        let gas_left = self.blockchain().get_gas_left();
+        let gas_left = self.blockchain().get_gas_left() / 2;
```
sasurobert, line 39: why do you divide by 2 ? it should work without it

sc-locked-asset-factory/src/locked_asset.rs
```diff
@@ -48,6 +54,43 @@ pub trait LockedAssetModule: asset::AssetModule {
...
+                accept_funds_func.as_slice(),
+                self.transfer_exec_gas_limit().get(),
+            ),
+            OptionalArg::None => {
+                let no_func: &[u8] = &[];
+                (no_func, 0u64)
+            }
+        };
+
+        let result = self.send().direct_esdt_nft_execute(
```
sasurobert, line 76: you should not care about the result when sending the tokens

dex/elrond_dex_pair/src/lib.rs
sasurobert: I understood that behind this function we actually have the transferAndExecute API so this should be ok. However you should not care about the results of one send.

sasurobert: If the transaction fails on the sender part - the execution will stop anyway and completely revert to previous state.

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -341,21 +362,114 @@ pub trait Farm: rewards::RewardsModule + config::ConfigModule {
...
+
+    fn send_tokens(
+        &self,
+        token: &TokenIdentifier,
+        nonce: Nonce,
+        amount: &Self::BigUint,
+        destination: &Address,
+        opt_accept_funds_func: &OptionalArg<BoxedBytes>,
+    ) -> SCResult<()> {
+        if nonce == 0 {
```
sasurobert, line 398: I think this if is duplicated - previously you already checked with_locked_rewards - thus on else you will have unlocked rewards -> thus you can call send_fft - as before.

sasurobert (changes requested):

dex/elrond_dex_router/src/factory.rs
catalinnnn: there are other operations happening after actual deploy, so a part of gas has to be reserved for them

sc-locked-asset-factory/src/locked_asset.rs
catalinnnn: Fixed

dex/elrond_dex_pair/src/lib.rs
catalinnnn: Fixed

distribution/sc-proxy-dex/src/proxy_common.rs
```diff
@@ -31,6 +35,76 @@ pub trait ProxyCommonModule {
...
     }
 
+    #[payable("*")]
+    #[endpoint]
+    fn acceptPay(
+        &self,
+        #[payment_token] token_id: TokenIdentifier,
+        #[payment] amount: Self::BigUint,
+    ) {
+        let tx_hash = self.blockchain().get_tx_hash();
```
sasurobert, line 45: nice resolve - this function always uses farm and swap contracts in a synchronous mode - thus there is no way for someone to rewrite these value - like the last tx-hash. Although please add tests which validate in all cases.

distribution/sc-proxy-dex/src/proxy_common.rs
```diff
@@ -31,6 +35,76 @@ pub trait ProxyCommonModule {
...
+        } else {
+            self.last_tx_hash().set(&tx_hash);
+        }
+
+        let token_nonce = self.call_value().esdt_token_nonce();
+        self.last_tx_accepted_funds()
+            .insert((token_id, token_nonce), amount);
+    }
+
+    fn validate_received_funds_on_current_tx(
```
sasurobert, line 62: I think you need one more verification - if I want to be picky. To verify that you did not get some funds which you did not want ? like you wanted exactly this token - and only that. none others. If there are 2 token IDs and you wanted only one - the tx should fail.

distribution/sc-proxy-dex/src/proxy_common.rs
```diff
@@ -31,6 +35,76 @@ pub trait ProxyCommonModule {
...
+    #[payable("*")]
+    #[endpoint]
+    fn acceptPay(
+        &self,
+        #[payment_token] token_id: TokenIdentifier,
+        #[payment] amount: Self::BigUint,
+    ) {
+        let tx_hash = self.blockchain().get_tx_hash();
+
+        if !self.last_tx_hash().is_empty() {
```
sasurobert, line 47: I do not know if you have to actually use the txHash. before calling for example exitFarm - you could set the storage of accepted tokens to nil, than you get the call of acceptPay - where you update the storage - and later you verify. It would save some gas. You would use one key for the storage mapper.

sasurobert (changes requested):

distribution/sc-proxy-dex/src/proxy_common.rs
catalinnnn: done

distribution/sc-proxy-dex/src/proxy_common.rs
catalinnnn: done

distribution/sc-proxy-dex/src/proxy_common.rs
catalinnnn: i did tests. they are on another branch and PR

sc-locked-asset-factory/src/locked_asset.rs
```diff
@@ -48,6 +54,36 @@ pub trait LockedAssetModule: asset::AssetModule {
         );
     }
 
+    fn send_tokens(
```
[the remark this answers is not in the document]
catalinnnn: no easy way to do this with the current rust framework. i'd say we move those functions after we have a proper way to do it clean

dex/elrond_dex_router/src/factory.rs
sasurobert: you will always get back the unused gas. so there is no need for / 2.

sc-locked-asset-factory/src/locked_asset.rs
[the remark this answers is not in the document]
sasurobert: alright.

dex/elrond_dex_router/src/factory.rs
catalinnnn: removed / 2

distribution/sc-proxy-dex/src/proxy_common.rs
```diff
@@ -31,6 +36,78 @@ pub trait ProxyCommonModule {
...
     }
 
+    #[payable("*")]
+    #[endpoint]
+    fn acceptPay(
+        &self,
+        #[payment_token] token_id: TokenIdentifier,
+        #[payment] amount: Self::BigUint,
+    ) {
+        if self.current_tx_accepted_funds().len() > MAX_FUNDS_EXTRIES {
```
sasurobert, line 46: why is this condition needed ?

distribution/sc-proxy-dex/src/proxy_farm.rs
```diff
@@ -160,6 +166,34 @@ pub trait ProxyFarmModule: proxy_common::ProxyCommonModule + proxy_pair::ProxyPa
         let farmed_token_returned = farm_result.0;
         let reward_token_returned = farm_result.1;
 
+        if reward_token_returned.token_id == farmed_token_returned.token_id {
```
sasurobert, line 169: please refactor lines 169 to 189 - we might find something easier. put into a separate function and refactor the if else structures.

distribution/sc-proxy-dex/src/proxy_common.rs
catalinnnn: because invalidation is only done in enter/exit/claimRewards endpoint, one might send funds just to increase the storage indefinitely

distribution/sc-proxy-dex/src/proxy_farm.rs
catalinnnn: fixed

sasurobert (approved):

PR #75, opened 2021-05-29
Audit fixing

catalinnnn:
Fix update last_reward_block_nonce at each step
Added check for fee configuration in pair sc
Added possibility for user to check his temporary funds in dex proxy sc

### DIFF
--- a/distribution/sc-proxy-dex/src/proxy_pair.rs
+++ b/distribution/sc-proxy-dex/src/proxy_pair.rs
@@ -590,7 +590,7 @@ pub trait ProxyPairModule: proxy_common::ProxyCommonModule {
         &self,
         address: &Address,
     ) -> MultiResultVec<GenericEsdtAmountPair<Self::BigUint>> {
-        let vec: Vec<GenericEsdtAmountPair<Self::BigUint>> = self
+        MultiResultVec::from_iter(self
             .temporary_funds(address)
             .iter()
             .map(|x| {
@@ -602,8 +602,8 @@ pub trait ProxyPairModule: proxy_common::ProxyCommonModule {
                     amount,
                 }
             })
-            .collect();
-        MultiResultVec::from_iter(vec)
+            .collect::<Vec<GenericEsdtAmountPair<Self::BigUint>>>()
+        )
     }
 
     #[storage_mapper("funds")]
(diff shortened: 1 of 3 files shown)


distribution/sc-proxy-dex/src/proxy_pair.rs
```diff
@@ -561,13 +586,27 @@ pub trait ProxyPairModule: proxy_common::ProxyCommonModule {
...
+        &self,
+        address: &Address,
+    ) -> MultiResultVec<GenericEsdtAmountPair<Self::BigUint>> {
+        let vec: Vec<GenericEsdtAmountPair<Self::BigUint>> = self
+            .temporary_funds(address)
+            .iter()
+            .map(|x| GenericEsdtAmountPair {
+                token_id: x.0 .0,
+                token_nonce: x.0 .1,
+                amount: x.1,
```
andrei-marinica, line 599: It's okay as it is but could be made a little easier to read like this
```suggestion
                let (key, amount) = x;
                let (token_id, token_nonce) = key;
                GenericEsdtAmountPair { token_id, token_nonce, amount }
```

andrei-marinica (approved):

distribution/sc-proxy-dex/src/proxy_pair.rs
catalinnnn: Didn't know about this syntax. Thank you. Fixed

distribution/sc-proxy-dex/src/proxy_pair.rs
```diff
@@ -561,13 +586,31 @@ pub trait ProxyPairModule: proxy_common::ProxyCommonModule {
...
+            .map(|x| {
+                let (key, amount) = x;
+                let (token_id, token_nonce) = key;
+                GenericEsdtAmountPair {
+                    token_id,
+                    token_nonce,
+                    amount,
+                }
+            })
+            .collect();
```
claudiu725, line 605: You can `.collect()` directly into a `MultiResultVec`, without needing to use an intermediate `Vec`

claudiu725 (changes requested):

distribution/sc-proxy-dex/src/proxy_pair.rs
catalinnnn: Done

sasurobert (approved):

PR #76, opened 2021-05-31
Small refactor

catalinnnn:
We already have the minimum value from above. So only one call would do it.

### DIFF
--- a/distribution/sc-proxy-dex/src/proxy_pair.rs
+++ b/distribution/sc-proxy-dex/src/proxy_pair.rs
@@ -374,8 +374,6 @@ pub trait ProxyPairModule: proxy_common::ProxyCommonModule {
             let difference = assets_received - locked_assets_invested.clone();
             self.send()
                 .direct(&caller, &asset_token_id, &difference, &[]);
-            self.send()
-                .esdt_local_burn(&asset_token_id, &locked_assets_invested);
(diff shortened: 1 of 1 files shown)


mateigraura (approved):

andrei-marinica (approved):

dorin-iancu (approved):

distribution/sc-proxy-dex/src/proxy_pair.rs
```diff
@@ -374,22 +374,18 @@ pub trait ProxyPairModule: proxy_common::ProxyCommonModule {
...
                 &difference,
             );
-            self.send()
-                .esdt_local_burn(&asset_token_id, &assets_received);
-        } else {
-            self.send()
-                .esdt_local_burn(&asset_token_id, &assets_received);
         }
 
+        self.send()
```
sasurobert, line 386: move burn before send tokens. have all these things everywhere before the actual send

catalinnnn: send and burn operations are the last operations in the contracts. their order does not quite matter since no calls like get_esdt_balance are done within the contracts

claudiu725 (approved):

sasurobert (approved):

PR #93, opened 2021-06-15
Audit fixing

catalinnnn:


distribution/sc-distribution-rs/src/lib.rs
```diff
@@ -265,21 +282,26 @@ pub trait Distribution: asset::AssetModule + global_op::GlobalOperationModule {
...
+            if self.blockchain().get_gas_left() < search_gas_limit {
                 break;
             }
+
             if lower <= user_asset_key.spread_epoch && user_asset_key.spread_epoch <= higher {
                 to_remove_keys.push(user_asset_key);
             }
         }
 
+        let mut removed = 0;
```
dorin-iancu, line 295: Instead of keeping an amount of removed, might be better to save the length before entering the "for" and returning old_length - current_length

sc-locked-asset-factory/src/lib.rs
```diff
@@ -43,15 +50,17 @@ pub trait LockedAssetFactory:
     fn whitelist(&self, address: Address) -> SCResult<()> {
         only_owner!(self, "Permission denied");
 
-        self.whitelisted_contracts().insert(address);
+        let is_new = self.whitelisted_contracts().insert(address);
+        require!(is_new, "Address already whitelisted");
```
dorin-iancu, line 54: There's no reason to return error if the address was already in the whitelist, as it does not harm :)

sc-locked-asset-factory/src/locked_asset.rs
```diff
@@ -175,9 +177,7 @@ pub trait LockedAssetModule: asset::AssetModule {
             percents_sum += milestone.unlock_percent;
         }
 
-        if !unlock_milestones.is_empty() {
-            require!(percents_sum == 100, "Percents do not sum up to 100");
-        }
+        require!(percents_sum == 100, "Percents do not sum up to 100");
```
dorin-iancu, line 180: Extract the "100" as a PERCENTAGE_TOTAL constant or similar.

dorin-iancu (changes requested):

sc-locked-asset-factory/src/lib.rs
catalinnnn: I know, but it seems it helps RV with their modeling

distribution/sc-distribution-rs/src/lib.rs
catalinnnn: Done

sc-locked-asset-factory/src/locked_asset.rs
catalinnnn: Done

dorin-iancu (approved):

PR #94, opened 2021-06-15
Feat/swap a to c

catalinnnn:


dex/elrond_dex_pair/src/lib.rs
```diff
@@ -357,15 +362,15 @@ pub trait Pair:
     }
 
     #[payable("*")]
-    #[endpoint(swapTokensFixedInput)]
-    fn swap_tokens_fixed_input(
+    #[endpoint]
+    fn swapTokensFixedInput(
```
dorin-iancu, line 366: Any reason for the rename? Rust standard requires functions to have snake_case naming.

dex/elrond_dex_pair/src/lib.rs
```diff
@@ -428,19 +433,22 @@ pub trait Pair:
...
+        Ok(FftTokenAmountPair {
+            token_id: token_out,
+            amount: amount_out_optimal,
+        })
     }
 
     #[payable("*")]
-    #[endpoint(swapTokensFixedOutput)]
-    fn swap_tokens_fixed_output(
+    #[endpoint]
```
dorin-iancu, line 443: Same question as above.

dex/elrond_dex_router/src/factory.rs
```diff
@@ -132,21 +134,82 @@ pub trait FactoryModule {
         }
     }
 
+    #[view(getPair)]
+    fn get_pair(
+        &self,
+        first_token_id: TokenIdentifier,
+        second_token_id: TokenIdentifier,
+    ) -> Address {
```
dorin-iancu, line 142: I think it's better to return `Option<Address>` instead of constantly checking for address == Address::zero()

claudiulataretu (approved):

dex/elrond_dex_router/src/pair_manager.rs
```diff
@@ -0,0 +1,198 @@
...
+        Ok(())
+    }
+
+    #[payable("*")]
+    #[endpoint]
+    fn acceptPay(&self) {}
+
+    #[payable("*")]
+    #[endpoint]
+    fn multiPairSwap(
```
dorin-iancu, line 60: naming

dex/elrond_dex_router/src/pair_manager.rs
```diff
@@ -0,0 +1,198 @@
...
+                last_received_token_id = token_amount_out.token_id;
+                last_received_amount = token_amount_out.amount;
+                residuum_vec.push(residuum);
+            } else {
+                return sc_error!("Invalid function to call");
+            }
+        }
+
+        while !residuum_vec.is_empty() {
+            let residuum = residuum_vec.pop().unwrap();
```
dorin-iancu, line 111: `unwrap` generates "panic" code even if this is always valid. Please use `unwrap_or_default` or `unwrap_or_else` instead.

dex/elrond_dex_router/src/util.rs
```diff
@@ -0,0 +1,90 @@
...
+                    accept_funds_func.as_slice(),
+                    self.transfer_exec_gas_limit().get(),
+                ),
+                OptionalArg::None => {
+                    let no_func: &[u8] = &[];
+                    (no_func, 0u64)
+                }
+            };
+
+            let result = self.send().direct_esdt_execute(
```
dorin-iancu, line 44: Rework this so it doesn't make a call towards an unexisting function if OptionalArg is None.

dex/elrond_dex_router/src/util.rs
```diff
@@ -0,0 +1,90 @@
...
+                destination,
+                token,
+                amount,
+                gas_limit,
+                function,
+                &ArgBuffer::new(),
+            );
+
+            match result {
```
dorin-iancu, line 53: I think you can use "?" operator on `Result<T>`.

dex/elrond_dex_router/src/util.rs
```diff
@@ -0,0 +1,90 @@
...
+                Result::Err(_) => {
+                    sc_error!("Direct esdt nft execute failed")
+                }
+            }
+        } else {
+            Ok(())
+        }
+    }
+
+    #[endpoint]
```
dorin-iancu, line 64: #[endpoint(setTransferExecGasLimit)]

dorin-iancu (changes requested):

dex/elrond_dex_pair/src/lib.rs
catalinnnn: After long debugging sessions while rust framework was not really stable and released, I've encountered the situation where an exec on dest would not work if the function was defined with snake_case and the endpoint with camelCase. Probably now it's fine, will test and update

dex/elrond_dex_router/src/util.rs
catalinnnn: What do you mean an unexisting function? If OptionalArg is None, then the function is passes as empty and gas is 0.

dorin-iancu: Oh, I see now. Fair enough.

dex/elrond_dex_router/src/pair_manager.rs
catalinnnn: How can it generate panic code when there's a check for empty value above?

dex/elrond_dex_router/src/util.rs
catalinnnn: done

dex/elrond_dex_router/src/util.rs
catalinnnn: done

dex/elrond_dex_router/src/pair_manager.rs
catalinnnn: done

dex/elrond_dex_router/src/pair_manager.rs
catalinnnn: done

dex/elrond_dex_router/src/factory.rs
catalinnnn: done

dex/elrond_dex_pair/src/lib.rs
catalinnnn: done

dex/elrond_dex_pair/src/lib.rs
catalinnnn: done

dorin-iancu (approved):

PR #95, opened 2021-06-16
Front-running protection

catalinnnn:
Implementation for: https://ethresear.ch/t/improving-front-running-resistance-of-x-y-k-market-makers/1281


dex/elrond_dex_pair/src/fee.rs
```diff
@@ -108,6 +108,11 @@ pub trait FeeModule:
         let mut reserve = self.pair_reserve(token).get();
         reserve += amount;
         self.pair_reserve(token).set(&reserve);
+
+        let mut virtual_reserve = self.pair_virtual_reserve(token, token).get();
```
dorin-iancu, line 112: Can use the `update` method instead of get -> modify -> set. Something like this:

```
self.pair_virtual_reserve(token, token).update(|reserve|  *reserve += amount);
```

This makes it so they storage key is built only once instead of twice.

dex/elrond_dex_pair/src/lib.rs
```diff
@@ -429,10 +428,18 @@ pub trait Pair:
...
-        self.update_reserves(&reserve_token_in, &reserve_token_out, &token_in, &token_out);
+        self.set_virtual_reserves(
+            &token_in,
+            &token_in,
+            &token_out,
+            &reserve_token_in,
+            &reserve_token_out,
+        );
+        self.increase_token_reserve(&token_in, &amount_in_after_fee);
+        self.try_decrease_token_reserve(&token_out, &amount_out_optimal)?;
```
dorin-iancu, line 439: Optional: This doesn't really matter, but it might be better to do the possibly error-returning call before the call that always works. (i.e. try decrease token_out, and then increase token_id)

dex/elrond_dex_pair/src/lib.rs
```diff
@@ -504,10 +512,18 @@ pub trait Pair:
...
-        self.update_reserves(&reserve_token_in, &reserve_token_out, &token_in, &token_out);
+        self.set_virtual_reserves(
+            &token_in,
+            &token_in,
+            &token_out,
+            &reserve_token_in,
+            &reserve_token_out,
+        );
+        self.increase_token_reserve(&token_in, &amount_in_optimal_after_fee);
+        self.try_decrease_token_reserve(&token_out, &amount_out)?;
```
dorin-iancu, line 523: Optional: order.

dex/elrond_dex_pair/src/liquidity_pool.rs
```diff
@@ -5,6 +5,8 @@ use super::amm;
 use super::config;
 use dex_common::FftTokenAmountPair;
 
+type Nonce = u64;
```
dorin-iancu, line 8: Maybe extract type aliases into a "util" crate, if you have such a thing in this project. Makes it so you don't have to keep defining in every file.

dex/elrond_dex_pair/src/liquidity_pool.rs
```diff
@@ -146,17 +148,52 @@ pub trait LiquidityPoolModule: amm::AmmModule + config::ConfigModule {
...
+        amount: &Self::BigUint,
+    ) -> SCResult<()> {
+        let old_value = self.pair_reserve(token_id).get();
+        require!(&old_value > amount, "Not enough reserves");
+        self.pair_reserve(token_id).set(&(&old_value - amount));
+        Ok(())
+    }
+
+    fn decrease_token_reserve(&self, token_id: &TokenIdentifier, amount: &Self::BigUint) {
+        let old_value = self.pair_reserve(token_id).get();
```
dorin-iancu, line 193: Use `update()` instead.

dex/elrond_dex_pair/src/liquidity_pool.rs
```diff
@@ -146,17 +148,52 @@ pub trait LiquidityPoolModule: amm::AmmModule + config::ConfigModule {
...
+        second_token_reserve: &Self::BigUint,
+    ) {
+        self.pair_virtual_reserve(token_side_id, first_token)
+            .set(first_token_reserve);
+        self.pair_virtual_reserve(token_side_id, second_token)
+            .set(second_token_reserve);
+    }
+
+    fn increase_token_reserve(&self, token_id: &TokenIdentifier, amount: &Self::BigUint) {
+        let new_value = &self.pair_reserve(token_id).get() + amount;
```
dorin-iancu, line 177: `update()`

dex/elrond_dex_pair/src/liquidity_pool.rs
```diff
@@ -146,17 +148,52 @@ pub trait LiquidityPoolModule: amm::AmmModule + config::ConfigModule {
...
+        self.pair_reserve(token_id).set(&new_value);
+    }
+
+    fn try_decrease_token_reserve(
+        &self,
+        token_id: &TokenIdentifier,
+        amount: &Self::BigUint,
+    ) -> SCResult<()> {
+        let old_value = self.pair_reserve(token_id).get();
```
dorin-iancu, line 186:
```
self.pair_reserve(token_id)
            .update(|reserve| {
                require!(&*reserve > amount, "Not enough reserves");
                *reserve -= amount;
                Ok(())
            })
```

dorin-iancu (changes requested):

dex/elrond_dex_pair/src/fee.rs
catalinnnn: Didn't know about update(). Thanks!
Done

dex/elrond_dex_pair/src/lib.rs
catalinnnn: Done

dex/elrond_dex_pair/src/lib.rs
catalinnnn: Done

dex/elrond_dex_pair/src/liquidity_pool.rs
catalinnnn: Done

dex/elrond_dex_pair/src/liquidity_pool.rs
catalinnnn: Done

dex/elrond_dex_pair/src/liquidity_pool.rs
catalinnnn: Done

dex/elrond_dex_pair/src/liquidity_pool.rs
catalinnnn: Done

dorin-iancu (approved):

PR #101, opened 2021-06-24
Reclaim Endpoint in Proxy

catalinnnn:
Reclaim will now work with one token at a time

### DIFF
--- a/distribution/sc-proxy-dex/src/proxy_pair.rs
+++ b/distribution/sc-proxy-dex/src/proxy_pair.rs
@@ -73,39 +73,21 @@ pub trait ProxyPairModule: proxy_common::ProxyCommonModule {
     #[endpoint(reclaimTemporaryFundsProxy)]
     fn reclaim_temporary_funds_proxy(
         &self,
-        first_token_id: TokenIdentifier,
-        first_token_nonce: Nonce,
-        second_token_id: TokenIdentifier,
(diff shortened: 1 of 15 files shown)


distribution/sc-proxy-dex/src/proxy_pair.rs
```diff
@@ -73,39 +73,21 @@ pub trait ProxyPairModule: proxy_common::ProxyCommonModule {
...
         let caller = self.blockchain().get_caller();
-        let first_token_amount = self
-            .temporary_funds(&caller)
-            .get(&(first_token_id.clone(), first_token_nonce))
-            .unwrap_or_else(Self::BigUint::zero);
-        let second_token_amount = self
+        let token_amount = self
             .temporary_funds(&caller)
-            .get(&(second_token_id.clone(), second_token_nonce))
+            .get(&(token_id.clone(), token_nonce))
```
dorin-iancu, line 82: Since you're using the same tuple twice (here and on line 85), consider saving it as a variable. This makes it so you also don't have to clone the token_id twice.

dorin-iancu (approved):

distribution/sc-proxy-dex/src/proxy_pair.rs
catalinnnn: done

PR #102, opened 2021-06-26
Dev/compound rewards

catalinnnn:


dex/elrond_dex_farm/src/lib.rs
```diff
@@ -343,22 +368,92 @@ pub trait Farm:
...
         &self,
         farming_token_id: &TokenIdentifier,
-        farming_amount: &mut Self::BigUint,
-        apr_multiplier: u8,
+        farming_amount: &Self::BigUint,
         destination: &Address,
-        with_locked_rewards: bool,
         opt_accept_funds_func: &OptionalArg<BoxedBytes>,
     ) -> SCResult<()> {
-        if with_locked_rewards {
```
sasurobert, line 355: why is this code missing in the new implementation?

sasurobert (dismissed):

dex/elrond_dex_farm/src/lib.rs
catalinnnn: before, the farming token amount one used to enter a farm were either 1:1 or 1:2 with the farm token amount. now, there is no longer the case and we use a variable in the token attributes so we know exactly the amount one used to enter farm. so no need for other calculations here since we already have the value we want to send back

dex/elrond_dex_farm/src/farm_token_merge.rs
```diff
@@ -0,0 +1,180 @@
...
+        let deposit_len = self.nft_deposit(caller).len();
+        require!(deposit_len != 0 || replic.is_some(), "No tokens to merge");
+
+        let mut index = 1;
+        let mut tokens = Vec::new();
+        let farm_token_id = self.farm_token_id().get();
+
+        while index <= deposit_len {
+            let entry = self.nft_deposit(caller).get(index);
+            require!(entry.token_id == farm_token_id, "Not a farm token");
```
virgil-serbanuta, line 66: Some of these entries may be empty after calling `withdraw_token_from_deposit`, would it be worth skipping over them?

virgil-serbanuta (commented):
One more batch of comments...

dex/elrond_dex_farm/src/farm_token.rs
```diff
@@ -0,0 +1,187 @@
...
+        token_nonce: u64,
+    ) -> SCResult<FarmTokenAttributes<Self::BigUint>> {
+        let token_info = self.blockchain().get_esdt_token_data(
+            &self.blockchain().get_sc_address(),
+            token_id,
+            token_nonce,
+        );
+
+        let farm_attributes = token_info.decode_attributes::<FarmTokenAttributes<Self::BigUint>>();
+        match farm_attributes {
```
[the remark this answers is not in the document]
catalinnnn: Yes but would imply having panic!("error message") in the code and I'm not a fan of that

common/modules/nft_deposit/src/nft_deposit.rs
```diff
@@ -0,0 +1,145 @@
...
+                "Deposit is full"
+            );
+            self.nft_deposit(&caller).push(&payment);
+        }
+
+        Ok(())
+    }
+
+    #[endpoint(withdrawAllTokensFromDeposit)]
+    fn withdraw_all_tokens_from_deposit(
```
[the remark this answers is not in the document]
catalinnnn: It actually don't, just checked. Though it works like a regular vector from Rust, apparently not. Will fix this

common/modules/token_merge/src/token_merge.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn weighted_average(&self, dataset: Vec<ValueWeight<Self::BigUint>>) -> Self::BigUint {
+        let mut weight_sum = Self::BigUint::zero();
+        dataset
+            .iter()
+            .for_each(|x| weight_sum = &weight_sum + &x.weight);
+
+        let mut elem_weight_sum = Self::BigUint::zero();
+        dataset
+            .iter()
+            .for_each(|x| elem_weight_sum = &weight_sum + &(&x.value * &x.weight));
```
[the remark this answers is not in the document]
catalinnnn: Yes. Good catch. Will fix this

common/modules/nft_deposit/src/nft_deposit.rs
```diff
@@ -0,0 +1,145 @@
...
+                entry_updated = true;
+                break;
+            }
+
+            index += 1;
+        }
+
+        if !entry_updated {
+            require!(
+                deposit_len + 1 < self.nft_deposit_max_len().get(),
```
yzhang90, line 50: why `<` instead of `<=`? or it doesn't matter?

common/modules/nft_deposit/src/nft_deposit.rs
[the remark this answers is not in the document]
yzhang90: what does this mean? the `clear_entry` function will not modify the length of the vector?

dex/elrond_dex_farm/src/farm_token_merge.rs
```diff
@@ -0,0 +1,180 @@
...
+        Ok(first_elem_value)
+    }
+
+    fn aggregated_with_lock_rewards(&self, tokens: &[FarmToken<Self::BigUint>]) -> SCResult<bool> {
+        let first_elem_value = tokens.get(1).unwrap().attributes.with_locked_rewards;
+        let mut same_value = true;
+        tokens
+            .iter()
+            .for_each(|x| same_value &= first_elem_value == x.attributes.with_locked_rewards);
+        require!(same_value, "Cannot compute with locked rewards aggregate");
```
virgil-serbanuta, line 134: Why is it OK not to handle the case when there are tokens with different with_locked_rewards values?

common/modules/nft_deposit/src/nft_deposit.rs
```diff
@@ -0,0 +1,145 @@
...
+        opt_accept_funds_func: &OptionalArg<BoxedBytes>,
+    ) -> SCResult<()> {
+        require!(
+            index >= 1 && index <= self.nft_deposit(caller).len(),
+            "Out of range index"
+        );
+
+        let entry = self.nft_deposit(caller).get(index);
+        self.nft_deposit(caller).clear_entry(index);
+        self.send_nft_tokens(
```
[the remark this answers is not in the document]
catalinnnn: Fixed. No more empty entries

common/modules/nft_deposit/src/nft_deposit.rs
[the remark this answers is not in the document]
catalinnnn: Yes, it will just leave an empty entry and will not modify the vec len. Replaced VecMapper with SingleValueMapper of Vec so we'll no longer have empty entries

dex/elrond_dex_farm/src/farm_token_merge.rs
catalinnnn: Fixed. No more empty entries

common/modules/nft_deposit/src/nft_deposit.rs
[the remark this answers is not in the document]
catalinnnn: Replaced with <=. Was < because the deposit of max 10 would be merged with one more token that was not yet created and will result in a 11 tokens merge, but that shouldn't be a problem

virgil-serbanuta (commented):
Next batch of comments. Sorry if some of them do not make sense anymore.

sc-locked-asset-factory/src/locked_asset_token_merge.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn merge_and_send_tokens(
+        &self,
+        #[var_args] opt_accept_funds_func: OptionalArg<BoxedBytes>,
+    ) -> SCResult<GenericEsdtAmountPair<Self::BigUint>> {
+        let caller = self.blockchain().get_caller();
+        let (amount, attrs) = self.get_merged_locked_asset_token_amount_and_attributes(&caller)?;
+        let locked_asset_token = self.locked_asset_token_id().get();
+
+        self.burn_deposit_tokens(&caller);
+        self.nft_deposit(&caller).clear();
```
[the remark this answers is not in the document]
catalinnnn: Did this in last commit

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -511,183 +592,40 @@ pub trait Farm:
...
-        amount: &Self::BigUint,
+        #[payment_token] payment_token_id: TokenIdentifier,
+        #[payment_nonce] payment_token_nonce: Nonce,
+        #[payment_amount] payment_amount: Self::BigUint,
     ) -> SCResult<()> {
-        let farm_amount = self.get_farm_token_supply();
-        require!(&farm_amount >= amount, "Not enough supply");
-        self.nft_burn_tokens(farm_token_id, farm_token_nonce, amount);
-        Ok(())
+        self.deposit_tokens(payment_token_id, payment_token_nonce, payment_amount)
```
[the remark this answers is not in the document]
catalinnnn: deposit_tokens function does this check

dex/elrond_dex_farm/src/farm_token_merge.rs
catalinnnn: If you merge some tokens with different with_locked_rewards values and someone claims the rewards would mean you have to send some rewards and some locked_rewards. In order to keep things -not that complicated- I decided to avoid that case.

common/modules/nft_deposit/src/nft_deposit.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved

common/modules/nft_deposit/src/nft_deposit.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved

common/modules/token_merge/src/token_merge.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved

dex/elrond_dex_farm/src/farm_token.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved

dex/elrond_dex_farm/src/farm_token_merge.rs
virgil-serbanuta: Resolved

sc-locked-asset-factory/src/locked_asset_token_merge.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved

dex/elrond_dex_farm/src/lib.rs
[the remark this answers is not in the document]
virgil-serbanuta: Ah, it's the `self.nft_deposit_accepted_token_ids().contains(&payment_token_id)` check, right? If so, then please mark my comment as resolved.

dex/elrond_dex_farm/src/farm_token_merge.rs
virgil-serbanuta: Ok, thanks, that answers my question, but I would leave this as "unresolved" to remind me to check the implications of this require failing.

common/modules/nft_deposit/src/nft_deposit.rs
```diff
@@ -0,0 +1,133 @@
...
+            &opt_accept_funds_func,
+        );
+
+        Ok(())
+    }
+
+    fn burn_deposit_tokens(
+        &self,
+        caller: &Address,
+        deposit: &[GenericEsdtAmountPair<Self::BigUint>],
```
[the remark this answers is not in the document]
catalinnnn: The argument is just a copy of the storage. Here the storage itself must be cleared

dex/elrond_dex_farm/src/lib.rs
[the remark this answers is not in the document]
catalinnnn: Yes

common/modules/token_send/src/token_send.rs
```diff
@@ -0,0 +1,71 @@
...
+            amount,
+            gas_limit,
+            function,
+            &ArgBuffer::new(),
+        );
+    }
+
+    #[view(getTransferExecGasLimit)]
+    #[storage_mapper("transfer_exec_gas_limit")]
+    fn transfer_exec_gas_limit(&self) -> SingleValueMapper<Self::Storage, u64>;
```
yzhang90, line 70: The gas limit for transfer esdt and nft will be the same?

common/modules/nft_deposit/src/nft_deposit.rs
```diff
@@ -0,0 +1,133 @@
...
+        address: &Address,
+    ) -> SingleValueMapper<Self::Storage, Vec<GenericEsdtAmountPair<Self::BigUint>>>;
+
+    #[view(getnftDepositMaxLen)]
+    #[storage_mapper("nft_deposit_max_len")]
+    fn nft_deposit_max_len(&self) -> SingleValueMapper<Self::Storage, usize>;
+
+    #[view(getNftDepositAcceptedTokenIds)]
+    #[storage_mapper("nft_deposit_accepted_token_ids")]
+    fn nft_deposit_accepted_token_ids(&self) -> SetMapper<Self::Storage, TokenIdentifier>;
```
yzhang90, line 132: The `nft_deposit_accepted_token_ids` has to be a set because it is used in the proxy contract, right? For the farm contract, it will be a singleton set.

dex/elrond_dex_farm/src/farm_token.rs
```diff
@@ -0,0 +1,187 @@
...
+        token_id: &TokenIdentifier,
+        token_nonce: u64,
+    ) -> SCResult<FarmTokenAttributes<Self::BigUint>> {
+        let token_info = self.blockchain().get_esdt_token_data(
+            &self.blockchain().get_sc_address(),
+            token_id,
+            token_nonce,
+        );
+
+        let farm_attributes = token_info.decode_attributes::<FarmTokenAttributes<Self::BigUint>>();
```
yzhang90, line 151: can we use `?` here to simplify the error handling?

common/modules/token_send/src/token_send.rs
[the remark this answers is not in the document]
catalinnnn: Yes

common/modules/nft_deposit/src/nft_deposit.rs
catalinnnn: Yes

dex/elrond_dex_farm/src/farm_token.rs
catalinnnn: Yes. Will fix

common/common_structs/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+pub type Epoch = u64;
+
+#[derive(TopEncode, TopDecode, PartialEq, TypeAbi)]
+pub struct FftTokenAmountPair<BigUint: BigUintApi> {
+    pub token_id: TokenIdentifier,
+    pub amount: BigUint,
+}
+
+#[derive(TopEncode, TopDecode, NestedEncode, NestedDecode, PartialEq, TypeAbi, Clone)]
+pub struct GenericEsdtAmountPair<BigUint: BigUintApi> {
```
yzhang90, line 16: Since it has `token_nonce` field, why call it GenericESDTAmountPair?

common/modules/nft_deposit/src/nft_deposit.rs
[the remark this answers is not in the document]
yzhang90: I think you did something like
```

File: dex/elrond_dex_farm/src/farm_token.rs (hunk 14 above)
Comment (catalinnnn) [reply dex/elrond_dex_farm/src/farm_token.rs:151]: Actually we can't. Seems weird but I did some debugging. Here's the reason: https://github.com/ElrondNetwork/elrond-wasm-rs/blob/master/elrond-wasm/src/types/io/sc_error.rs#L57. This has a method for EncodeError but does not have one for DecodeError sadly. We'll stick to the handling until this will be added in the rust framework.

Comment (yzhang90) [reply dex/elrond_dex_farm/src/farm_token.rs:151]: ok.

File: common/common_structs/src/lib.rs (hunk 15 above)
Comment (catalinnnn) [reply common/common_structs/src/lib.rs:16]: NFTs and SFTs are ESDTs too. I'll rename the struct to GenericTokenAmountPair to avoid confusions

File: common/modules/nft_deposit/src/nft_deposit.rs (hunk 11 above)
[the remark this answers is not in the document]
Comment (catalinnnn) [reply common/modules/nft_deposit/src/nft_deposit.rs:103]: Because reading a vector from storage might be gas expensive. I observed that every time when burning the tokens the deposit is already loaded, so passed it as parameter here to save some gas. It is a tradeoff between clean code and gas efficiency. I'd say the gas is more important in this case.

File: common/modules/nft_deposit/src/nft_deposit.rs (hunk 6 above)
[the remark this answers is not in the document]
Comment (yzhang90) [reply common/modules/nft_deposit/src/nft_deposit.rs]: ok

File: common/modules/nft_deposit/src/nft_deposit.rs (hunk 11 above)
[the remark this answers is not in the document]
Comment (yzhang90) [reply common/modules/nft_deposit/src/nft_deposit.rs]: ok

File: common/common_structs/src/lib.rs (hunk 15 above)
Comment (yzhang90) [reply common/common_structs/src/lib.rs]: ok. I thought ESDT is the ERC20 equivalent in Elrond. It is not the case?

Comment (catalinnnn) [reply common/common_structs/src/lib.rs]: Fungible ESDT tokens are the equivalent of ERC20 since they have just a type (ID) and an amount. Non fungible ESDT tokens also have metadata and nonce and I think their equivalent would be ERC721, although I'm not 100% sure

File: dex/elrond_dex_farm/src/lib.rs [hunk 16]
```diff
@@ -219,29 +233,35 @@ pub trait Farm:
         }
 
         let farming_token_id = self.farming_token_id().get();
-        let mut farming_token_amount = amount.clone();
+        let mut initial_farming_token_amount = &(&farm_attributes.initial_farming_amount * &amount)
```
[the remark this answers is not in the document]
catalinnnn: Indeed it's the same calculus. The reason I did not replace everywhere was to minimize the number of changed lines. But at a second thought it might improve readability. Will update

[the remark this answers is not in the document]
catalinnnn: done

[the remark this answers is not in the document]
virgil-serbanuta: Resolved

sc-locked-asset-factory/src/locked_asset_token_merge.rs
```diff
@@ -0,0 +1,156 @@
...
+            "Too many milestones"
+        );
+
+        let mut new_unlock_milestones = Vec::new();
+        unlock_epoch_amount_merged.iter().for_each(|x| {
+            if x.1 != Self::BigUint::zero() {
+                let unlock_percent = &(&x.1 * &Self::BigUint::from(100u64)) / &sum;
+                new_unlock_milestones.push(UnlockMilestone {
+                    unlock_epoch: x.0,
+                    unlock_percent: unlock_percent.to_u64().unwrap() as u8,
```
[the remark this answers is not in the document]
catalinnnn: Shouldn't be a problem but here's a limit on the unlock schedule vec len and the 0's might occupy those with no use. Added a check for this case.

sc-locked-asset-factory/src/locked_asset_token_merge.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+        require!(
+            unlock_epoch_amount_merged.len() < MAX_MILESTONES_IN_SCHEDULE,
+            "Too many milestones"
+        );
+
+        let mut new_unlock_milestones = Vec::new();
+        unlock_epoch_amount_merged.iter().for_each(|x| {
+            if x.1 != Self::BigUint::zero() {
+                let unlock_percent = &(&x.1 * &Self::BigUint::from(100u64)) / &sum;
```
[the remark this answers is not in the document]
catalinnnn: This is true. Should no be able to merge too highly fragmented amounts of intial amounts of locked asset tokens. Added a require! here.

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -219,29 +218,41 @@ pub trait Farm:
         }
 
         let farming_token_id = self.farming_token_id().get();
-        let mut farming_token_amount = amount.clone();
+        let mut initial_farming_token_amount = self.rule_of_three(
+            &amount,
+            &farm_attributes.current_farm_amount,
+            &farm_attributes.initial_farming_amount,
+        );
+        reward += self.rule_of_three(
```
yzhang90, line 226: why not `decrease_reward_reserve` on the compound rewards?

catalinnnn: the amount was already substracted from reward_reserve upon compoundRewards transaction

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -0,0 +1,332 @@
...
+        &self,
+        tokens: &[WrappedFarmToken<Self::BigUint>],
+    ) -> SCResult<GenericTokenAmountPair<Self::BigUint>> {
+        let mut wrapped_lp_tokens = Vec::new();
+
+        for token in tokens.iter() {
+            let wrapped_lp_token_amount = self.rule_of_three(
+                &token.token_amount.amount,
+                &token.attributes.farming_token_amount,
+                &token.attributes.farm_token_amount,
```
virgil-serbanuta, line 267: Should the last two arguments be swapped?
```
self.rule_of_three(
                &token.token_amount.amount,
                &token.attributes.farm_token_amount,
                &token.attributes.farming_token_amount,
)
```

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+
+    fn merge_wrapped_lp_tokens_from_farm(
+        &self,
+        tokens: &[WrappedFarmToken<Self::BigUint>],
+    ) -> SCResult<GenericTokenAmountPair<Self::BigUint>> {
+        let mut wrapped_lp_tokens = Vec::new();
+
+        for token in tokens.iter() {
```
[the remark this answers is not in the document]
catalinnnn: The check is for avoiding doing computation like average means and ops like execute on dest context. `merge_wrapped_lp_tokens_from_farm` has the same logic for 1 or more tokens. Notice `merge_wrapped_lp_tokens_from_farm` uses `merge_locked_asset_tokens_from_wrapped_lp` which has this check

sc-locked-asset-factory/src/locked_asset_token_merge.rs
```diff
@@ -0,0 +1,160 @@
...
+{
+    #[endpoint(mergeLockedAssetTokens)]
+    fn merge_locked_asset_tokens(
+        &self,
+        #[var_args] opt_accept_funds_func: OptionalArg<BoxedBytes>,
+    ) -> SCResult<GenericTokenAmountPair<Self::BigUint>> {
+        let caller = self.blockchain().get_caller();
+        let deposit = self.nft_deposit(&caller).get();
+        let (amount, attrs) = self.get_merged_locked_asset_token_amount_and_attributes(&deposit)?;
+        let locked_asset_token = self.locked_asset_token_id().get();
```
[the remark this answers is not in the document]
catalinnnn: Sure, will add the check there also

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
catalinnnn: Yes, fixed. Thank you for pointing out

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -298,16 +309,40 @@ pub trait Farm:
             self.decrease_reward_reserve(&reward)?;
         }
 
-        let new_attributes = FarmTokenAttributes {
+        let new_initial_farming_amount = self.rule_of_three(
+            &amount,
+            &farm_attributes.current_farm_amount,
+            &farm_attributes.initial_farming_amount,
+        );
+        let new_reward_amount = self.rule_of_three(
```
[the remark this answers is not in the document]
catalinnnn: done

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -219,29 +218,41 @@ pub trait Farm:
...
+            &amount,
+            &farm_attributes.current_farm_amount,
+            &farm_attributes.initial_farming_amount,
+        );
+        reward += self.rule_of_three(
+            &amount,
+            &farm_attributes.current_farm_amount,
+            &farm_attributes.compounded_reward,
+        );
+        require!(
```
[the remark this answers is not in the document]
catalinnnn: done

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
virgil-serbanuta: Resolved

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        Ok(GenericTokenAmountPair {
+            token_id: wrapped_lp_token_id,
+            token_nonce: new_nonce,
+            amount,
+        })
+    }
+
+    fn get_farm_attributes(
```
virgil-serbanuta, line 313: Is this used anywhere?

catalinnnn: No. Removed it

dex/elrond_dex_pair/src/config.rs
```diff
@@ -26,10 +26,6 @@ pub trait ConfigModule {
     #[storage_mapper("extern_swap_gas_limit")]
     fn extern_swap_gas_limit(&self) -> SingleValueMapper<Self::Storage, u64>;
 
-    #[view(getTranferExecGasLimit)]
```
yzhang90, line 29: why this view function is removed?

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -343,22 +378,127 @@ pub trait Farm:
...
+        );
+        let new_compound_reward_amount = self.rule_of_three(
+            &payment_amount,
+            &farm_attributes.current_farm_amount,
+            &farm_attributes.compounded_reward,
+        ) + reward;
+
+        let mut new_attributes = FarmTokenAttributes {
+            reward_per_share: current_rps,
+            entering_epoch: farm_attributes.entering_epoch,
```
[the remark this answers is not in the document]
catalinnnn: How's that? What value should be used instead?

dex/elrond_dex_pair/src/config.rs
catalinnnn: Some simplification in the framework functions. ExecuteOnDest context no longer needs gas limit and will use the gas available for the whole transaction

dex/elrond_dex_farm/src/lib.rs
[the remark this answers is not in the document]
catalinnnn: Sorry, misread your comment. No merge happening here. The entering epoch is the same for this poisition. Just claimed the rewards and reinvested them

sc-locked-asset-factory/src/locked_asset_token_merge.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved.

sc-locked-asset-factory/src/locked_asset_token_merge.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -0,0 +1,332 @@
...
+        tokens: &[WrappedFarmToken<Self::BigUint>],
+    ) -> SCResult<GenericTokenAmountPair<Self::BigUint>> {
+        let mut wrapped_lp_tokens = Vec::new();
+
+        for token in tokens.iter() {
+            let wrapped_lp_token_amount = self.rule_of_three(
+                &token.token_amount.amount,
+                &token.attributes.farm_token_amount,
+                &token.attributes.farming_token_amount,
+            );
```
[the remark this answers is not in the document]
catalinnnn: done

dex/elrond_dex_farm/src/lib.rs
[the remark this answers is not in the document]
yzhang90: ok. If  a user does the compound manually(ie. exit the farm and reenter the farm with rewards), then the entering epoch will be the new one. If a user uses the compound function, then the entering epoch will always be the old one. right?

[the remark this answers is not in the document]
catalinnnn: yes

distribution/sc-proxy-dex/src/wrapped_lp_token_merge.rs
```diff
@@ -0,0 +1,189 @@
...
+        &self,
+        tokens: &[WrappedLpToken<Self::BigUint>],
+    ) -> GenericTokenAmountPair<Self::BigUint> {
+        let locked_asset_factory_addr = self.locked_asset_factory_address().get();
+        let locked_asset_token = self.locked_asset_token_id().get();
+
+        if tokens.len() == 1 {
+            let token = tokens[0].clone();
+
+            let amount = self.rule_of_three(
```
virgil-serbanuta, line 145: I'm trying to figure out how this `rule_of_three` call works. As far as I can see (though I probably made a mistake somewhere, the code is a bit complex), this takes a fraction of `token.token_amount.amount/token.attributes.lp_token_total_amount`, which can be up to `2`, out of `token.attributes.locked_assets_invested`. Why does it make sense for the amount to be twice the value of `token.attributes.locked_assets_invested`?

Let's take an example:

1. The user adds `asset_amount` of `locked_asset_token_id()` as liquidity to a pair contract. This gets recorded in a `WrappedLpTokenAttributes` struct (with `lp_token_total_amount=lp_amount` and `token.attributes.locked_assets_invested=asset_amount`), and the user receives `lp_amount` of `wrapped_farm_token_id()`. Note that the relation between `asset_amount` and `lp_amount` is not easy to describe (they are not necessarily equal).
2. The user adds `lp_amount` of `wrapped_farm_token_id()` to a farm contract, with locked rewards. This gets recorded in a `WrappedFarmTokenAttributes` structure, and the user receives `2*lp_amount` of `wrapped_farm_token_id()`.
3. The user takes this `2*lp_amount` of `wrapped_farm_token_id()` and deposits them. In principle, the user may deposit other amounts, also, in which case we will reach the similar `rule_of_three` line below.
4. The user calls `merge_wrapped_farm_tokens`. We can reach this line with the following sequence of calls: `merge_wrapped_farm_tokens -> merge_wrapped_farm_tokens_and_send -> merge_farming_tokens -> merge_wrapped_lp_tokens_from_farm -> merge_locked_asset_tokens_from_wrapped_lp`
5. This line calls rule_of_three(2 * lp_amount, lp_amount, asset_amount), with a result of `2*asset_amount`.

Did I make any mistakes? If not, why does it make sense to have `amount = 2 * asset_amount`? I think that this gets recorded again as a `locked_assets_invested` field, on which the user can probably apply steps 3-5 again.

catalinnnn: The 2 * asset_amount breaks when using Locked Mex to farm at MEX Farm after Compunding Rewards. After doing this, the relation between locked assets invested and farm tokens received breaks, meaning wrapped farm tokens amount won't be equal with 2 * locked assets invested anymore.

catalinnnn: + the apr multiplier can change later. The requirement was such that it can be changed for the new positions at any time. And  Proxy SC having to rely on a setting inside the farm token attributes (so a token it cannot create or burn) doesn't seem right

dex/elrond_dex_pair/src/lib.rs
```diff
@@ -528,41 +507,14 @@ pub trait Pair:
...
-
-            match result {
-                Result::Ok(_) => Ok(()),
-                Result::Err(_) => {
-                    sc_error!("Direct esdt nft execute failed")
-                }
-            }
-        } else {
-            Ok(())
+            self.send_fft_tokens(token, amount, destination, opt_accept_funds_func);
```
yzhang90, line 512: The `send_tokens` function just call `send_fft_tokens`, why not use `send_fft_tokens` directly?

dex/elrond_dex_pair/src/lib.rs
```diff
@@ -733,11 +675,6 @@ pub trait Pair:
         self.state().get() == State::Active
     }
 
-    #[view]
```
yzhang90, line 736: there is no need to view the lp token id?

dex/elrond_dex_pair/src/lib.rs
catalinnnn: To avoid using the "if" multiple times

dex/elrond_dex_pair/src/lib.rs
catalinnnn: moved the function. it's still used

dex/elrond_dex_pair/src/lib.rs
yzhang90: I see. I missed that when reviewing the change.

dex/multi_token_farm/src/lib.rs
```diff
@@ -7,10 +7,7 @@ mod rewards;
 
 elrond_wasm::imports!();
 elrond_wasm::derive_imports!();
-use dex_common::{FftTokenAmountPair, GenericEsdtAmountPair};
-
-type Epoch = u64;
-type Nonce = u64;
+use common_structs::{Epoch, FftTokenAmountPair, GenericTokenAmountPair, Nonce};
```
yzhang90, line 10: Just to make sure: this file is deprecated and it is not under the scope of review

dex/multi_token_farm/src/rewards.rs
```diff
@@ -1,7 +1,7 @@
 elrond_wasm::imports!();
 elrond_wasm::derive_imports!();
 
-type Nonce = u64;
+use common_structs::Nonce;
```
yzhang90, line 4: Just to make sure: this file is deprecated and it is not under the scope of review

distribution/sc-proxy-dex/src/wrapped_lp_token_merge.rs
virgil-serbanuta: Thanks for explaining on Google Meet, my mistake was in steps 4-5: `merge_farming_tokens` would convert `2*lp_amount` to `lp_amount`, so the call in step 5 would use `lp_amount` instead of `2*lp_amount`.

Resolved.

dex/multi_token_farm/src/lib.rs
catalinnnn: Yes. No review needed.

dex/multi_token_farm/src/rewards.rs
catalinnnn: Yes

sc-locked-asset-factory/src/locked_asset_token_merge.rs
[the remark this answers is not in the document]
virgil-serbanuta: Resolved, I think.

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -0,0 +1,317 @@
...
+                .deposit_tokens(
+                    entry.attributes.farm_token_id.clone(),
+                    entry.attributes.farm_token_nonce,
+                    entry.token_amount.amount.clone(),
+                )
+                .execute_on_dest_context();
+        }
+
+        self.farm_contract_merge_proxy(farm_contract.clone())
+            .merge_farm_tokens(OptionalArg::Some(BoxedBytes::from(ACCEPT_PAY_FUNC_NAME)))
```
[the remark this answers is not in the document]
catalinnnn: The reason is that we've added a check on sender part for errors. The contracts are in the same shard and if the send fails, the execution should be terminated on the sender part and the whole transaction will be reverted. When we introduced this validate functions, we didn't really know what to use and this check was the strongest and should have worked with any method. This is the reason of validate functions. They do no harm since it's just a check and I don't want to delete them because of this. Also do not want to add another function for the gas. We'll like to make as little changes as possible because we are time restricted.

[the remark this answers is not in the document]
virgil-serbanuta: Resolved.

BTW, it was slightly easier to review the code with those checks, because sometimes one could see locally that certain properties hold (e.g. that the received amounts use a specific token). However, if those validations were not needed, it would not have been worth it to keep them just for that.

distribution/sc-proxy-dex/src/wrapped_lp_token_merge.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn get_merged_wrapped_lp_token_attributes(
+        &self,
+        tokens: &[WrappedLpToken<Self::BigUint>],
+        merged_locked_asset_token_amount: &GenericTokenAmountPair<Self::BigUint>,
+    ) -> WrappedLpTokenAttributes<Self::BigUint> {
+        let mut lp_token_amount = 0u64.into();
+
+        tokens
+            .iter()
+            .for_each(|x| lp_token_amount += &x.attributes.lp_token_total_amount);
```
yzhang90, line 134: the `lp_token_total_amount` is used with `locked_assets_invested` to calculate the locked tokens when removing the liquidity. So the total number of wrapped lp tokens generated and the `lp_token_total_amount` in the token attribute should always match. When merging wrapped_lp tokens, the `lp_token_total_amount` is the sum of all `lp_token_total_amount` from the tokens to merge. While the total number of generated wrapped_lp_tokens is the sum of all the tokens to merge. These two numbers can be different.

catalinnnn: indeed. the value calculated here ("lp_token_amount") was not right. fixed

yzhang90: @virgil-serbanuta please double check

distribution/sc-proxy-dex/src/wrapped_lp_token_merge.rs
```diff
@@ -0,0 +1,197 @@
...
+        &self,
+        tokens: &[WrappedLpToken<Self::BigUint>],
+    ) -> GenericTokenAmountPair<Self::BigUint> {
+        let locked_asset_factory_addr = self.locked_asset_factory_address().get();
+        let locked_asset_token = self.locked_asset_token_id().get();
+
+        if tokens.len() == 1 {
+            let token = tokens[0].clone();
+
+            let amount = self.rule_of_three(
```
yzhang90, line 153: check `amount` is greater than 0?

distribution/sc-proxy-dex/src/wrapped_lp_token_merge.rs
virgil-serbanuta: I think it's fine now.

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -0,0 +1,331 @@
...
+            };
+            wrapped_lp_tokens.push(wrapped_lp_token);
+        }
+
+        let merged_locked_token_amount =
+            self.merge_locked_asset_tokens_from_wrapped_lp(&wrapped_lp_tokens);
+        let merged_wrapped_lp_token_amount =
+            self.get_merged_wrapped_lp_tokens_amount(&wrapped_lp_tokens);
+        let lp_token_amount = FftTokenAmountPair {
+            token_id: wrapped_lp_tokens[0].attributes.lp_token_id.clone(),
```
virgil-serbanuta, line 302: Are all the `wrapped_lp_tokens[*].attributes.lp_token_id` values identical? If yes, why?

catalinnnn: Should have missed the check for that. Added

distribution/sc-proxy-dex/src/wrapped_lp_token_merge.rs
catalinnnn: Added checks for most rule of three

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
virgil-serbanuta: Where is the new check?

catalinnnn: Just 6 lines above

virgil-serbanuta: Ah, sorry, I'm not sure why I didn't see it before. I thought I `git pull`ed the last version, but maybe I didn't...

Resolved.

yzhang90: The function is finally used by `merge_farming_tokens`, before calling the function, we check all the farm_token_ids are identical. Does it imply that the lp tokens are identical?

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -0,0 +1,331 @@
...
+            new_nonce,
+            &merged_farm_token_amount.amount,
+            caller,
+            &opt_accept_funds_func,
+        )?;
+
+        Ok(())
+    }
+
+    fn require_deposit_empty_or_tokens_are_wrapped_farm_tokens(&self) -> SCResult<()> {
```
yzhang90, line 110: This function is not used anymore?

catalinnnn: It's used in 3 other places in proxy_farm.rs

distribution/sc-proxy-dex/src/proxy_farm.rs
```diff
@@ -252,60 +243,112 @@ pub trait ProxyFarmModule:
...
-            Result::Err(_) => {
-                return sc_error!("Decoding error");
-            }
-        }
-    }
+        let wrapped_farm_token_attrs =
+            self.get_wrapped_farm_token_attributes(&payment_token_id, payment_token_nonce)?;
+        let farm_token_id = wrapped_farm_token_attrs.farm_token_id;
+        let farm_token_nonce = wrapped_farm_token_attrs.farm_token_nonce;
+        let farm_amount = payment_amount.clone();
```
yzhang90, line 291: The compound function only works for the lockedMEX. Is it better to check if the farming_token_id = locked_asset_id here?

yzhang90, line 291: The compound function only works for lockedMEX token here. Shall we check it?

catalinnnn: It will fail anyways in the farm contract so it is safe. No need for another check

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -219,29 +212,37 @@ pub trait Farm:
...
                 self.burn_tokens(&reward_token_id, &penalty_amount);
                 reward -= penalty_amount;
             }
 
-            penalty_amount = self.get_penalty_amount(&farming_token_amount);
+            penalty_amount = self.get_penalty_amount(&initial_farming_token_amount);
             if penalty_amount > 0 {
                 self.burn_tokens(&farming_token_id, &penalty_amount);
-                farming_token_amount -= penalty_amount;
+                initial_farming_token_amount -= penalty_amount;
```
virgil-serbanuta, line 236: Why isn't there a call to `decrease_farming_token_reserve` here?

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -343,29 +365,130 @@ pub trait Farm:
...
+        )?;
+        let new_compound_reward_amount = self.rule_of_three(
+            &payment_amount,
+            &farm_attributes.current_farm_amount,
+            &farm_attributes.compounded_reward,
+        ) + reward;
+
+        let mut new_attributes = FarmTokenAttributes {
+            reward_per_share: current_rps,
+            entering_epoch: self.blockchain().get_block_epoch(),
```
yzhang90, line 421: I noticed that this field is `farm_attributes.entering_epoch` in the claimRewards. How about here?

catalinnnn: Here it should be like this. I'm not a fan of resetting in some instances and not resetting in others

yzhang90: So if a user compounds the rewards after a year and exits the farm. Then the all the rewards will be locked after a year?

catalinnnn: Pretty much yes

yzhang90: Humm. I wonder whether users will like this compound feature or not. Bascially, each time users compound their reward, they are delaying the time to unlock the compound rewards.

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -298,23 +300,43 @@ pub trait Farm:
...
+        let new_nonce = self.create_farm_tokens_by_merging(
+            &mut farm_amount,
+            &farm_token_id,
+            &mut new_attributes,
+            &caller,
+        )?;
         self.send_nft_tokens(
             &farm_token_id,
             new_nonce,
             &amount,
```
virgil-serbanuta, line 336: Shouldn't this be sending `farm_amount` (i.e. the merged amount) instead of `amount`?

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -343,29 +365,130 @@ pub trait Farm:
...
+            &payment_amount,
+            &current_rps,
+            &farm_attributes.reward_per_share,
+        );
+
+        if reward > 0 {
+            self.decrease_reward_reserve(&reward)?;
+        }
+
+        let farm_token_id = self.farm_token_id().get();
```
virgil-serbanuta, line 404: I think that farm_token_id already has this value (see line 381)

dex/elrond_dex_farm/src/lib.rs
virgil-serbanuta: FWIW, to my untrained eye it seems that it could have been a merge between `(payment_amount, farm_attributes.entering_epoch)` and `(reward*multiplier, get_block_epoch)`, similar to what would have happened if you would generate a new farm token for `reward*multiplier`, then you merged that with the existing entry.

dex/elrond_dex_farm/src/lib.rs
catalinnnn: True, but since it's already fixed in this branch: https://github.com/ElrondNetwork/sc-dex-rs/pull/106 we'll fix it when we merge this one too

dex/elrond_dex_farm/src/lib.rs
catalinnnn: Shouldn't be a problem

dex/elrond_dex_farm/src/lib.rs
catalinnnn: Merging with the existing entry changed similary

dex/elrond_dex_farm/src/lib.rs
catalinnnn: added

sc-locked-asset-factory/src/locked_asset_token_merge.rs
```diff
@@ -0,0 +1,162 @@
...
+        tokens.iter().for_each(|locked_token| {
+            locked_token
+                .attributes
+                .unlock_schedule
+                .unlock_milestones
+                .iter()
+                .for_each(|milestone| {
+                    unlock_epoch_amount.push((
+                        milestone.unlock_epoch,
+                        self.rule_of_three(
```
yzhang90, line 106: should we use rule_of_three_non_zero() here?

catalinnnn: Zero values should not be a problem as long as not all of them are zero

dex/elrond_dex_farm/src/lib.rs
virgil-serbanuta: Resolved

dex/elrond_dex_farm/src/farm_token_merge.rs
virgil-serbanuta: Resolved

distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
virgil-serbanuta: Not sure what @yzhang90 thinks about this, but, as far as I am concerned, this is resolved: line 294 calls self.require_wrapped_lp_tokens_from_same_pair(&wrapped_lp_tokens)?; , which checks that the lp_tokens are identical.

Of course, depending on the answer to Yi's question, that require might not have been needed.

dex/elrond_dex_farm/src/lib.rs
virgil-serbanuta: Resolved

virgil-serbanuta:
With the exception of this comment: https://github.com/ElrondNetwork/sc-dex-rs/pull/102#discussion_r678667451 , this looks good to me. Given that the PR referenced there ( #106 ) is non-trivial, it may have its own issues, so I'm not sure if I should take it as a fix or not given that I did not review it. On the other hand, you probably understand #106 together with the context for this PR much better than me, so do whatever makes sense to you.

Of course, I could always check more things, but, since you mentioned that you'd like to finish by the end of this week, I guess that this is a good point to stop my review.

claudiulataretu (dismissed):

claudiulataretu (approved):

PR #110, opened 2021-07-23
fix

catalinnnn:


distribution/sc-proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -255,7 +256,7 @@ pub trait WrappedFarmTokenMerge:
         let locked_asset_token_id = self.locked_asset_token_id().get();
 
         if farming_token_id == locked_asset_token_id {
-            Ok(self.merge_locked_asset_tokens_from_wrapped_farm(tokens))
+            self.merge_locked_asset_tokens_from_wrapped_farm(tokens)
```
claudiulataretu, line 259: the function returns a SCResult. Don't you need to return an Ok() for this?

catalinnnn: merge_locked_asset_tokens_from_wrapped_farm returns just that

PR #113, opened 2021-07-29
buyback and burn mex tokens instead of burn LP

catalinnnn:


mateigraura (approved):

dex/elrond_dex_pair/src/lib.rs
```diff
@@ -317,6 +317,56 @@ pub trait Pair:
...
+        require!(
+            token_in == self.lp_token_identifier().get(),
+            "Wrong liquidity token"
+        );
+
+        let first_token_id = self.first_token_id().get();
+        let second_token_id = self.second_token_id().get();
+
+        let first_token_min_amount = 1u64.into();
+        let second_token_min_amount = 1u64.into();
```
claudiulataretu, line 341: You can use the get equivalent for liquidity view function from pair to determine these values.

dex/elrond_dex_pair/src/lib.rs
```diff
@@ -317,6 +317,56 @@ pub trait Pair:
             .into())
     }
 
+    #[payable("*")]
+    #[endpoint(removeLiquidityAndBuyBackAndBurnToken)]
+    fn remove_liquidity_and_burn_token(
```
claudiulataretu, line 322: Would be an option to enforce that only the farms can call this endpoint? Or a whitelisted address? I'm thinking what would happen if someone, by any chance, would call this method

dex/elrond_dex_pair/src/lib.rs
catalinnnn: What benefit would that bring?

dex/elrond_dex_pair/src/lib.rs
catalinnnn: Not a fan of making the setup even harder. The function has a pretty intuitive name

claudiulataretu (approved):

PR #119, opened 2021-08-16
Feat/swap a to c

catalinnnn:


dex/elrond_dex_router/src/pair_manager.rs
```diff
@@ -0,0 +1,206 @@
...
+                    pair_address,
+                    last_received_token_id,
+                    last_received_amount,
+                    token_wanted,
+                    amount_wanted,
+                );
+                last_received_token_id = token_amount_out.token_id;
+                last_received_amount = token_amount_out.amount;
+            } else if function == BoxedBytes::from(SWAP_TOKENS_FIXED_OUTPUT_FUNC_NAME) {
+                let (token_amount_out, residuum) = self.actual_swap_fixed_output(
```
claudiulataretu, line 100: The problem that I foresee here is that a user would end up with some dust from a token he doesn't want. Although it would be quite hard to do, we might need to think of a solution to have this dust at least in the first input token. Does UniswapV2 handles this residuum?

catalinnnn: I don't see a problem here. Having extra tokens is not something bad. Let's not overcomplicate. If he strictly does not want it, he can only swapTokensFixedInput and that's it. Uniswap does not handle any residuum I think, although it could be double checked.

claudiulataretu (approved):

dex/elrond_dex_router/src/pair_manager.rs
```diff
@@ -0,0 +1,206 @@
...
+use elrond_dex_pair::config::ProxyTrait as _;
+use elrond_dex_pair::fee::ProxyTrait as _;
+
+#[elrond_wasm::module]
+pub trait PairManagerModule:
+    state::StateModule + factory::FactoryModule + token_send::TokenSendModule
+{
+    #[only_owner]
+    #[endpoint(setFeeOn)]
+    fn set_fee_on(
```
sasurobert, line 25: what is set_fee_on and set_fee_off ?

sasurobert (approved):

PR #130, opened 2021-09-03
Feat/sft liquidity

catalinnnn:


common/modules/reserve_queue/src/reserve_queue.rs
```diff
@@ -0,0 +1,417 @@
...
+        queues: &mut Vec<u64>,
+        queue_id_opt: Option<u64>,
+    ) {
+        let queue_id = queue_id_opt.unwrap_or_else(|| self.get_queue_id(amount, precision));
+        let head = self.head(token_id, queue_id).get();
+
+        if head == 0 {
+            self.init_queue(token_id, queue_id, nonce, amount);
+            queues.push(queue_id);
+            queues.sort();
```
sasurobert, line 299: why do you sort when only one token ?

common/modules/reserve_queue/src/reserve_queue.rs
```diff
@@ -0,0 +1,417 @@
...
+        } else {
+            self.push_to_queue(token_id, queue_id, nonce, amount);
```
sasurobert, line 301: are you sorting on every push ?

common/modules/reserve_queue/src/reserve_queue.rs
```diff
@@ -0,0 +1,417 @@
...
+        &self,
+        token_id: &TokenIdentifier,
+        amount: &Self::BigUint,
+        precision: u64,
+        queues: &mut Vec<u64>,
+        queue_id: u64,
+    ) -> SCResult<Vec<NonceAmountPair<Self::BigUint>>> {
+        let is_fungible = !self.reserve(token_id, 0).is_empty();
+
+        if is_fungible {
```
sasurobert, line 82: do we actually need this if's and else's ? the queue should have actually only 1 element with fungible tokens. Which might still happen on Semi Fungible that everyone will put only 1 type of token inside - like one nonce ?

common/modules/reserve_queue/src/reserve_queue.rs
sasurobert: so when we push - the position remains the same ?

common/modules/reserve_queue/src/reserve_queue.rs
```diff
@@ -0,0 +1,417 @@
...
+        let mut queue_id = 0;
+
+        let precision_biguint = Self::BigUint::from(precision);
+        let mut amount_clone = amount.clone();
+        while amount_clone != 0 {
+            amount_clone /= &precision_biguint;
+            queue_id += 1;
+        }
+
+        queue_id
```
sasurobert, line 281: Is this the binary search ? which sets the actual position for one token-nonce pair ?

common/modules/reserve_queue/src/reserve_queue.rs
```diff
@@ -0,0 +1,417 @@
...
+        token_id: &TokenIdentifier,
+        nonce: Nonce,
+        amount: &Self::BigUint,
+        precision: u64,
+        queues: &mut Vec<u64>,
+    ) {
+        let mut elem = self.reserve(token_id, nonce).get();
+        let queue = elem.queue_id;
+        elem.amount += amount;
+        let new_queue_id = self.get_queue_id(&elem.amount, precision);
```
sasurobert, line 316: so these pushes would keep the list always somewhat sorted ? biggest values are always first - which should be enough, not ?

common/modules/reserve_queue/src/reserve_queue.rs
catalinnnn: Its a vector of queue ids. when a new queue is instantiated, it needs to be added to the vector, but also the vector has to remain sorted. It should not have more than say 30 elements, so should be pretty fast and cheap

common/modules/reserve_queue/src/reserve_queue.rs
catalinnnn: When pushing a new element to an existing queue, only its tail will be changed (and the last elem pointer to next)

common/modules/reserve_queue/src/reserve_queue.rs
catalinnnn: I used nonce 0 in order to know whether the head/tail elements are instantiated or not. Similar to NULL. May refactor this, but some magic value will be used for this check anyway

common/modules/reserve_queue/src/reserve_queue.rs
catalinnnn: This just maps an amount (which can be whatever) to a bucket (a queue). Element with X amount will be added to a queue with respect to a configured precision

frick1977:
Not able to claim MEX and no to sign staking and wapping signature. Total failure for me
Maiar Chrome

PR #187, opened 2021-10-13
0.20 Multitransfer

catalinnnn:


common/modules/token_send/src/token_send.rs
```diff
@@ -12,22 +55,22 @@ pub trait TokenSendModule {
...
 
         SCResult::from_result(self.raw_vm_api().direct_esdt_execute(
             destination,
             token,
             amount,
             gas_limit,
-            &function,
+            &ManagedBuffer::managed_from(self.type_manager(), function),
             &ManagedArgBuffer::new_empty(self.type_manager()),
         ))
```
dorin-iancu, line 75: Can use `.into()` here, I believe. Same in the other functions.

common/modules/token_send/src/token_send.rs
```diff
@@ -54,11 +97,129 @@ pub trait TokenSendModule {
...
+                    &payment.amount,
+                    &ManagedAddress::managed_from(self.type_manager(), destination),
+                    opt_accept_funds_func,
+                )
+            }
+        } else if len > 1 {
+            self.send_multiple_tokens(&compact_payments, destination, opt_accept_funds_func)
+        } else {
+            Ok(())
+        }
```
dorin-iancu, line 176: I think this can be made slightly cleaner with a match
```
match compact_payments.len() {
0 => 
1 => 
_ =>
```

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -108,33 +104,38 @@ pub trait Farm:
...
     ) -> SCResult<EnterFarmResultType<Self::Api>> {
         require!(self.is_active(), "Not active");
         require!(!self.farm_token_id().is_empty(), "No issued farm token");
+
+        let payments = self
+            .raw_vm_api()
+            .get_all_esdt_transfers()
+            .into_iter()
+            .collect::<Vec<EsdtTokenPayment<Self::Api>>>();
+        require!(payments.len() >= 1, "empty payments");
```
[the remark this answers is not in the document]
claudiulataretu: I would keep all the payments sent to this endpoint and use them to consolidate into a single position. We can agree on a maximum of farm tokens to be sent for merging to not exceed the gas limit

distribution/sc-proxy-dex/src/proxy_farm.rs
```diff
@@ -153,17 +150,23 @@ pub trait ProxyFarmModule:
...
         self.require_is_intermediated_farm(farm_address)?;
         self.require_wrapped_farm_token_id_not_empty()?;
         self.require_wrapped_lp_token_id_not_empty()?;
+
+        let payments = self
+            .raw_vm_api()
+            .get_all_esdt_transfers()
+            .into_iter()
+            .collect::<Vec<EsdtTokenPayment<Self::Api>>>();
+        require!(payments.len() == 1, "bad payment len");
```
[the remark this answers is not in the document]
catalinnnn: fixed

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -302,13 +308,22 @@ pub trait Farm:
...
     ) -> SCResult<ClaimRewardsResultType<Self::Api>> {
         require!(self.is_active(), "Not active");
         require!(!self.farm_token_id().is_empty(), "No issued farm token");
+
+        let payments = self
+            .raw_vm_api()
+            .get_all_esdt_transfers()
+            .into_iter()
+            .collect::<Vec<EsdtTokenPayment<Self::Api>>>();
+        require!(payments.len() >= 1, "bad payment len");
```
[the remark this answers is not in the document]
catalinnnn: they are not, refactored so it is clearer

dex/elrond_dex_farm/src/lib.rs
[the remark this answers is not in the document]
catalinnnn: they are not, refactored so it is clearer

common/modules/token_send/src/token_send.rs
```diff
@@ -3,49 +3,91 @@
 elrond_wasm::imports!();
 elrond_wasm::derive_imports!();
 
+extern "C" {
+    fn transferESDTNFTExecute(
```
sasurobert, line 7: why don't you implement the one with managed buffers ? managedMultiTransferESDTNFTExecute for example ?

common/modules/token_send/src/token_send.rs
```diff
@@ -54,11 +96,126 @@ pub trait TokenSendModule {
...
+                accept_funds_func.as_slice(),
+                self.transfer_exec_gas_limit().get(),
+            ),
+            OptionalArg::None => {
+                let no_func: &[u8] = &[];
+                (no_func, 0u64)
+            }
+        };
+
+        SCResult::from_result(self.raw_vm_api().direct_multi_esdt_transfer_execute(
```
sasurobert, line 121: can't you use this instead of defining the C API ?

common/modules/token_send/src/token_send.rs
```diff
@@ -54,11 +96,126 @@ pub trait TokenSendModule {
...
+                });
+
+                match existing_index {
+                    Some(index) => compact_payments[index].amount += &payment.amount,
+                    None => compact_payments.push(payment.clone()),
+                }
+            }
+        }
+
+        match compact_payments.len() {
```
sasurobert, line 151: we should not have code like this. Simplu use send_multiple_tokens everywhere

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -406,12 +417,20 @@ pub trait Farm:
...
         &self,
-        #[payment_token] payment_token_id: TokenIdentifier,
-        #[payment_amount] payment_amount: BigUint,
-        #[payment_nonce] payment_token_nonce: Nonce,
-        #[var_args] opt_accept_funds_func: OptionalArg<ManagedBuffer>,
+        #[var_args] opt_accept_funds_func: OptionalArg<BoxedBytes>,
     ) -> SCResult<CompoundRewardsResultType<Self::Api>> {
         require!(self.is_active(), "Not active");
+
+        let payments = self
```
sasurobert, line 424: take lines 424-433 into  separate function which does you all the conversion.

dex/elrond_dex_farm/src/lib.rs
```diff
@@ -476,13 +495,13 @@ pub trait Farm:
             &new_farm_contribution,
             &farm_token_id,
             &new_attributes,
-            &caller,
+            &payments[1..],
```
sasurobert, line 498: why do you start from 1 ?

distribution/sc-proxy-dex/src/proxy_farm.rs
```diff
@@ -133,6 +130,7 @@ pub trait ProxyFarmModule:
                 &farm_token_total_amount,
                 &farm_address,
                 &caller,
+                &payments[1..],
```
sasurobert, line 133: Why don't you rewrite this by sending all payments inside ?

sasurobert (approved):

common/modules/token_send/src/token_send.rs
catalinnnn: it will not work due to wasm compile error

common/modules/token_send/src/token_send.rs
catalinnnn: indeed it is much cleaner with managedMultiTransferESDTNFTExecute, thanks for the suggestion

claudiulataretu (approved):

dex/elrond_dex_farm/src/lib.rs
sasurobert: Isn't it better to refactor all this? Like making a different merge function? Which gets all the payments, merges them, burns them and creates the new one?

PR #191, opened 2021-10-19
cleanup: dex contracts

catalinnnn:
Removed "elrond" from files names
Removed FfftTokenAmount and GenericTokenAmount. Used EsdtTokenPayment provided by framework.
Removed send methods, kept just MultiESDTTransfer
Renamed folders so they contain less underscores
Updated tests
Removed old multi-token-farm
Removed old deprecated tests

### DIFF
--- a/common/modules/token_send/src/token_send.rs
+++ b/common/modules/token_send/src/token_send.rs
@@ -105,15 +105,7 @@ pub trait TokenSendModule {
             .collect()
     }
 
-    fn fungible_payment(
-        &self,
-        token_id: &TokenIdentifier,
-        amount: &BigUint,
-    ) -> EsdtTokenPayment<Self::Api> {
-        EsdtTokenPayment::from(token_id.clone(), 0, amount.clone())
-    }
-
-    fn nonfungible_payment(
+    fn create_payment(
         &self,
         token_id: &TokenIdentifier,
         nonce: u64,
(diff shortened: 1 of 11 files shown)


common/modules/token_send/src/token_send.rs
```diff
@@ -159,6 +105,22 @@ pub trait TokenSendModule {
...
 
+    fn fungible_payment(
+        &self,
+        token_id: &TokenIdentifier,
+        amount: &BigUint,
+    ) -> EsdtTokenPayment<Self::Api> {
+        EsdtTokenPayment::from(token_id.clone(), 0, amount.clone())
+    }
+
+    fn nonfungible_payment(
```
sasurobert, line 116: we could simply have one function to create - and receives nonce always.

dex/pair/src/lib.rs
```diff
@@ -154,50 +151,31 @@ pub trait Pair:
         let lp_token_id = self.lp_token_identifier().get();
         self.mint_tokens(&lp_token_id, &liquidity);
 
+        let mut payments = Vec::new();
+        payments.push(self.fungible_payment(&lp_token_id, &liquidity));
+        payments.push(self.fungible_payment(&expected_first_token_id, &first_token_unused));
+        payments.push(self.fungible_payment(&expected_second_token_id, &second_token_unused));
```
sasurobert, line 157: would rename to create_payment

catalinnnn: makes sense. done

claudiulataretu (approved):

PR #202, opened 2021-11-05
added even more checks

catalinnnn:


claudiulataretu (approved):

locked-asset/proxy-dex/src/wrapped_farm_token_merge.rs
```diff
@@ -162,6 +166,13 @@ pub trait WrappedFarmTokenMerge:
         &self,
         tokens: &[WrappedFarmToken<Self::Api>],
     ) -> SCResult<EsdtTokenPayment<Self::Api>> {
+        require!(!tokens.is_empty(), "tokens empty");
+        let locked_asset_token = self.locked_asset_token_id().get();
+        require!(
```
yzhang90, line 171: `require_wrapped_farm_tokens_have_same_token_ids` ensures that all the wrapped farm tokens have the same farm_token and farming_token. This is what I'm talking about in slack.

You already check this `farming_token_id == locked_asset_token` in the end of the function `merge_farming_tokens`. So it is no need to check it again here.

yzhang90: Otherwise, it looks good to me

PR #221, opened 2021-11-30
farm: add whitelisting addresses for farm endpoints

claudiulataretu:
- Address whitelisting is used for farm accepting LKMEX as farming token
where proxy will be whitelisted in farm contract

Signed-off-by: Claudiu Ion Lataretu <claudiu.lataretu@gmail.com>

### DIFF
--- a/dex/farm/src/lib.rs
+++ b/dex/farm/src/lib.rs
@@ -112,6 +112,8 @@ pub trait Farm:
         &self,
         #[var_args] opt_accept_funds_func: OptionalArg<ManagedBuffer>,
     ) -> SCResult<EnterFarmResultType<Self::Api>> {
+        let apr_multiplier = self.locked_rewards_apr_multiplier().get();
+        require!(apr_multiplier > 1, "farm does not support locked rewards");
         self.enter_farm_common(true, opt_accept_funds_func)
     }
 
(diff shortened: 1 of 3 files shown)


dex/farm/src/lib.rs
```diff
@@ -210,6 +221,15 @@ pub trait Farm:
         require!(payment_token_id == farm_token_id, "Bad input token");
         require!(amount > 0, "Payment amount cannot be zero");
 
+        let caller = self.blockchain().get_caller();
```
sasurobert, line 224: you have to make one function out of lines 224-232. And use a require for it. one liner for this endpoint.

dex/farm/src/lib.rs
```diff
@@ -112,6 +112,8 @@ pub trait Farm:
         &self,
         #[var_args] opt_accept_funds_func: OptionalArg<ManagedBuffer>,
     ) -> SCResult<EnterFarmResultType<Self::Api>> {
+        let apr_multiplier = self.locked_rewards_apr_multiplier().get();
+        require!(apr_multiplier > 1, "farm does not support locked rewards");
```
sasurobert, line 116: I think that it is better to make a completely new set of contracts - one which generates locked rewards - thus metaESDTs and one which generates with ESDTs. That will make the overall contracts smaller.

dex/farm/src/lib.rs
sasurobert: I would make a separate version for this contract. And would have especially for this case. You can start a new feat/branch - where you say farm with whitelist.

catalinnnn (changes requested):
Looks good, but needs testing, we cannot keep adding changes without adding mandos tests

PR #222, opened 2021-12-01
Remove check for init epoch for create and forward lkmex

claudiulataretu:
Signed-off-by: Claudiu Ion Lataretu <claudiu.lataretu@gmail.com>


catalinnnn (changes requested):
Dex v1_1 is already in prod. Might want to change the base branch to dex-v1_2

PR #227, opened 2021-12-06
whitelist

dorin-iancu:


sasurobert (commented):
move to a feature branch. And push the optimizations / without the whitelist to v1_2_split PR as well

PR #228, opened 2021-12-06
feat whitelist

dorin-iancu:


dex/farm/src/lib.rs
```diff
@@ -339,6 +343,9 @@ pub trait Farm:
     ) -> SCResult<CompoundRewardsResultType<Self::Api>> {
         require!(self.is_active(), "Not active");
 
+        let caller = self.blockchain().get_caller();
```
claudiulataretu, line 346: Compound does not apply to this type of farm contract. This could be removed if we are aiming for a smaller code base.

dex/farm/src/custom_rewards.rs
```diff
@@ -8,25 +8,33 @@ pub trait CustomRewardsModule:
...
         let current_block_nonce = self.blockchain().get_block_nonce();
         let last_reward_nonce = self.last_reward_block_nonce().get();
 
         if current_block_nonce > last_reward_nonce {
             let to_mint = self.calculate_per_block_rewards(current_block_nonce, last_reward_nonce);
 
-            if to_mint != 0 {
-                self.send().esdt_local_mint(token_id, 0, &to_mint);
-            }
+            // rewards are not minted, but rather deposited by the owner
```
claudiulataretu, line 27: I would put a check for the number of actual remained rewards from sc balance to be grater than the minted rewards for the current_block - last_block period. When we will get to 0, then the SC should itself call end_produce_rewards. Otherwise, any claim rewards transactions would fail with no rewards to send

dex/farm/src/lib.rs
sasurobert: agreed. good call.

dex/farm/src/lib.rs
```diff
@@ -82,7 +80,6 @@ pub trait Farm:
         self.owner().set(&self.blockchain().get_caller());
         self.reward_token_id().set(&reward_token_id);
         self.farming_token_id().set(&farming_token_id);
-        self.pair_contract_address().set(&pair_contract_address);
```
virgil-serbanuta, line 85: Since this is not initialized anymore, would it make sense to also delete the field?

dorin-iancu: Done, now it's only in the farm-with-lock contract.

sasurobert (approved):
This looks good.

catalinnnn:
Tests are failing since only one address can use the farm contract now. Tests are written based on the old style. Feel free to modify them or change them accordingly.

catalinnnn (changes requested):
Tests are failing since only one address can use the farm contract now. Tests are written based on the old style. Feel free to modify them or change them accordingly.

PR #232, opened 2021-12-14
Pair SC Contexts refactor

catalinnnn:
Pair Contracts Context refactor


dex/pair/src/contexts/ctx_helper.rs
```diff
@@ -0,0 +1,272 @@
...
+        context.set_output_payments(payments);
+    }
+
+    fn execute_output_payments(&self, context: &dyn Context<Self::Api>) {
+        self.send_multiple_tokens_if_not_zero(
+            context.get_caller(),
+            context.get_output_payments(),
+            context.get_opt_accept_funds_func(),
+        )
+        .unwrap();
```
dorin-iancu, line 196: Consider using the `assert!` macros instead of `unwrap()`. I think a signal error directly is more efficient than the `panic!` code of `unwrap()`.

dex/pair/src/contexts/add_liquidity.rs
```diff
@@ -0,0 +1,300 @@
...
+        payment_opt: &Option<&EsdtTokenPayment<M>>,
+    ) -> bool {
+        match payment_opt {
+            Some(payment) => min <= &payment.amount,
+            None => false,
+        }
+    }
+}
+
+impl<M: ManagedTypeApi> AddLiquidityContext<M> {
```
dorin-iancu, line 230: Consider `#[inline]`-ing some of the getter/setter methods. Same for the other contexts.

dex/pair/src/contexts/swap.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    pub fn get_reserve_in(&self) -> &BigUint<M> {
+        let payment_token_id = &self.tx_input.payments.input.token_identifier;
+
+        if payment_token_id == &self.storage_cache.first_token_id {
+            &self.storage_cache.first_token_reserve
+        } else if payment_token_id == &self.storage_cache.second_token_id {
+            &self.storage_cache.second_token_reserve
+        } else {
+            unreachable!()
```
dorin-iancu, line 237: Remove `unreachable!()`, as all it does it basically `panic!()` if that code section is ever reached. It doesn't actually remove the branch from the code, so you actually end up with even more code. If the code is unreachable anyway, simply remove the second `if` statement and return `&self.storage_cache.second_token_reserve` directly.

If you're unsure that's always the case for whatever reason, rewrite the method to either use `assert!()` or return an `Option<T>` instead. This applies to the other methods as well.

dex/pair/src/fee.rs
```diff
@@ -30,19 +33,17 @@ pub trait FeeModule:
...
     #[endpoint(whitelist)]
-    fn whitelist_endpoint(&self, address: ManagedAddress) -> SCResult<()> {
-        self.require_permissions()?;
+    fn whitelist_endpoint(&self, address: ManagedAddress) {
+        self.require_permissions();
         let is_new = self.whitelist().insert(address);
-        require!(is_new, "ManagedAddress already whitelisted");
-        Ok(())
+        assert!(self, is_new, ERROR_ALREADY_WHITELISTED);
     }
```
dorin-iancu, line 40: I'd say simply insert and don't return an error if the user was already whitelisted, as you're not changing anything in storage anyway, so nothing to revert.

dex/pair/src/fee.rs
```diff
@@ -30,19 +33,17 @@ pub trait FeeModule:
...
     #[endpoint(removeWhitelist)]
-    fn remove_whitelist(&self, address: ManagedAddress) -> SCResult<()> {
-        self.require_permissions()?;
+    fn remove_whitelist(&self, address: ManagedAddress) {
+        self.require_permissions();
         let is_removed = self.whitelist().remove(&address);
-        require!(is_removed, "ManagedAddresss not whitelisted");
-        Ok(())
+        assert!(self, is_removed, ERROR_NOT_WHITELISTED);
     }
```
dorin-iancu, line 47: Same here, simply remove, as trying to remove an element that does not exist does nothing.

dex/pair/src/fee.rs
```diff
@@ -51,25 +52,24 @@ pub trait FeeModule:
         pair_address: ManagedAddress,
         first_token: TokenIdentifier,
         second_token: TokenIdentifier,
-    ) -> SCResult<()> {
```
dorin-iancu, line 54: For methods that are deeply nested and only do very basic things, removing the `SCResult<()>` to not propagate it all over is fine. I don't think we should remove `SCResult` from _everywhere_ though, as that makes the code less clear and it doesn't really bring any improvement. I'd say keep endpoints that do checks themselves return SCResult<()> instead of using `assert!` everyhwere.

This also means keeping high-level check methods like `require_permissions` with `SCResult` return.

dex/pair/src/fee.rs
```diff
@@ -82,128 +82,86 @@ pub trait FeeModule:
...
+        if fee_amount == &0u64 {
             return;
         }
 
         let slices = self.destination_map().len() as u64;
-        if slices == 0 {
-            self.reinject(fee_token, fee_amount);
+        if slices == 0u64 {
             return;
         }
```
dorin-iancu, line 102: `if self.destination_map().is_empty() { return; }` is a bit cleaner.

dex/pair/src/fee.rs
```diff
@@ -256,32 +217,27 @@ pub trait FeeModule:
...
+        self.raw_vm_api()
+            .direct_esdt_execute(
+                &pair_address,
+                available_token,
+                available_amount,
+                self.extern_swap_gas_limit().get(),
+                &ManagedBuffer::from(SWAP_NO_FEE_AND_FORWARD_FUNC_NAME),
+                &arg_buffer,
+            )
+            .unwrap();
```
dorin-iancu, line 236: I suggest using a proxy for this, even if written by hand. Much easier to maintain.

dorin-iancu (changes requested):

dex/pair/src/config.rs
```diff
@@ -1,6 +1,20 @@
...
+macro_rules! assert {
+    ($self:expr, $cond:expr, $msg:expr $(,)?) => {
+        if !$cond {
+            assert!($self, $msg)
+        }
+    };
+    ($self:expr, $msg:expr $(,)?) => {
+        $self.raw_vm_api().signal_error($msg)
+    };
+}
```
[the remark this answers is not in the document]
catalinnnn: you need it since you call "self.raw_vm[...]" they dont catch the environment like in c/c++ the example. it's called macro hygiene

common/common_structs/src/lib.rs
```diff
@@ -14,6 +14,12 @@ pub struct TokenPair<M: ManagedTypeApi> {
     pub second_token: TokenIdentifier<M>,
 }
 
+impl<M: ManagedTypeApi> TokenPair<M> {
+    pub fn equals(&self, other: &TokenPair<M>) -> bool {
+        self.first_token == other.first_token && self.second_token == other.second_token
+    }
+}
+
```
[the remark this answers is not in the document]
catalinnnn: Somehow, it does not compile. After some debugging, i came to the conclusion that implementing it is not a big deal

dex/pair/src/fee.rs
catalinnnn: RV Suggested this behavior is better

dex/pair/src/fee.rs
catalinnnn: well, i use slice below, so i dont see a reason why using .is_empty() and .len() instead of just .len()

dex/pair/src/fee.rs
catalinnnn: Imo it's pretty straight forward what require_permissions does without returning a SCResult. Assert is just fine. Think that's why Rust guys implemented it

dex/pair/src/fee.rs
catalinnnn: RV Suggested this behavior is better

dex/pair/src/contexts/swap.rs
catalinnnn: unreachable -> assert is the intended behavior here

dex/pair/src/fee.rs
catalinnnn: It adds quite a bit of bytecode to the wasm file but ok

dex/pair/src/contexts/add_liquidity.rs
catalinnnn: done

dex/pair/src/contexts/ctx_helper.rs
catalinnnn: done

dex/pair/src/liquidity_pool.rs
```diff
@@ -10,148 +15,163 @@ const MINIMUM_LIQUIDITY: u64 = 1_000;
...
-        first_token_amount: BigUint,
-        second_token_amount: BigUint,
-    ) -> SCResult<BigUint> {
-        let first_token = self.first_token_id().get();
-        let second_token = self.second_token_id().get();
-        let total_supply = self.get_total_lp_token_supply();
-        let mut first_token_reserve = self.pair_reserve(&first_token).get();
-        let mut second_token_reserve = self.pair_reserve(&second_token).get();
+    fn pool_add_liquidity(&self, context: &mut AddLiquidityContext<Self::Api>) {
+        let zero = &BigUint::zero();
```
sasurobert, line 19: can we define zero as constant ?

dex/pair/src/liquidity_pool.rs
```diff
@@ -10,148 +15,163 @@ const MINIMUM_LIQUIDITY: u64 = 1_000;
...
         let mut liquidity: BigUint;
 
-        if total_supply == 0 {
-            liquidity = core::cmp::min(first_token_amount.clone(), second_token_amount.clone());
-            let minimum_liquidity = BigUint::from(MINIMUM_LIQUIDITY);
-            require!(
-                liquidity > minimum_liquidity,
-                "First tokens needs to be greater than minimum liquidity"
+        if context.get_lp_token_supply() == zero {
```
sasurobert, line 22: can't we do a separate function - like addInitialLiquidity - maybe protected at only_owner. and we can simply delete this if / else.

dex/pair/src/liquidity_pool.rs
catalinnnn: at the moment, no, i suggested something similar to Andrei already, we is working on it

dex/pair/src/config.rs
```diff
@@ -104,4 +112,19 @@ pub trait ConfigModule: token_send::TokenSendModule {
...
     fn lp_token_identifier(&self) -> SingleValueMapper<TokenIdentifier>;
+
+    #[view(getFirstTokenId)]
+    #[storage_mapper("first_token_id")]
+    fn first_token_id(&self) -> SingleValueMapper<TokenIdentifier>;
+
+    #[view(getSecondTokenId)]
+    #[storage_mapper("second_token_id")]
+    fn second_token_id(&self) -> SingleValueMapper<TokenIdentifier>;
+
```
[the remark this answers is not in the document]
catalinnnn: there's already a function for this and i think we should keep it for backwards compatibility
```
    #[view(getTotalSupply)]
    fn get_total_lp_token_supply(&self) -> BigUint {
        self.lp_token_supply().get()
    }
```

dex/pair/src/liquidity_pool.rs
catalinnnn: we could but... would be easy from the SC point of view, but i think it'll complicate as we go further on the sw stack, service and frontend

dex/pair/src/contexts/add_liquidity.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+}
+
+impl<M: ManagedTypeApi> AddLiquidityPayments<M> {
+    fn is_valid_payment(&self, payment_opt: &Option<&EsdtTokenPayment<M>>) -> bool {
+        match payment_opt {
+            Some(payment) => {
+                payment.amount != 0
+                    && payment.token_nonce == 0
+                    && payment.token_identifier.is_esdt()
```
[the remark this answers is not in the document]
catalinnnn: i had an idea that it was possible to send 0. anyway, would leave the check as it is, after the adding of bigUint::zero as a constant, the comparison should be really cheap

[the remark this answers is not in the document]
sasurobert: there is no way to get 0 as payment. It fails at protocol level.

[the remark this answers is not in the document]
sasurobert: I would not add redundant checks

dex/pair/src/liquidity_pool.rs
sasurobert: I do not think so. The first liquidity is added always from command line - not from frontend. I think it is safer to have a separate endpoint. It is part of the setup.

sasurobert: And when we want to integrate the initial price discovery mechanism - it is mandatory to have a whitelisted address and a secure endpoint.

dex/pair/src/contexts/add_liquidity.rs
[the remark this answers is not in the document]
catalinnnn: the check will stay.

dex/pair/src/liquidity_pool.rs
catalinnnn: will do on a separate PR :+1:

sasurobert (approved):

dorin-iancu (approved):

PR #237, opened 2022-01-04
Price Discovery SC

dorin-iancu:


price-discovery/src/create_pool.rs
```diff
@@ -0,0 +1,103 @@
...
+        let (lp_token, _, _) = contract_call.execute_on_dest_context().into_tuple();
+        self.lp_token_id().set(&lp_token.token_identifier);
+        self.total_lp_tokens_received().set(&lp_token.amount);
+
+        Ok(())
+    }
+
+    #[payable("*")]
+    #[endpoint]
+    fn accept_funds_func(&self) -> SCResult<()> {
```
sasurobert, line 76: maybe add a check that this endpoint is called only once.

price-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        self.accepted_token_id().set(&accepted_token_id);
+        self.start_epoch().set(&start_epoch);
+        self.end_epoch().set(&end_epoch);
+
+        Ok(())
+    }
+
+    #[only_owner]
+    #[payable("*")]
+    #[endpoint(depositInitialTokens)]
```
sasurobert, line 63: why is this needed ? like initial depoist ?

sasurobert: and why only owner ?

price-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        let caller = self.blockchain().get_caller();
+        self.mint_and_send_redeem_token(&caller, LAUNCHED_TOKEN_REDEEM_NONCE, &payment_amount);
+
+        Ok(())
+    }
+
+    #[payable("*")]
+    #[endpoint]
+    fn deposit(
```
sasurobert, line 87: I think only one deposit is enough. And according to tokenID you go for the tokenA or tokenB nonce (initial token or launchpad token).

price-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        &self,
+        #[payment_token] payment_token: TokenIdentifier,
+        #[payment_nonce] payment_nonce: u64,
+        #[payment_amount] payment_amount: BigUint,
+    ) -> SCResult<()> {
+        self.require_active()?;
+
+        let redeem_token_id = self.redeem_token_id().get();
+        require!(
+            payment_token == redeem_token_id && payment_nonce == ACCEPTED_TOKEN_REDEEM_NONCE,
```
sasurobert, line 115: withdraw can be done for accepted_token and launchpad_token as well. Some users have tokenA, others have tokenB they can simply enter with either of them. As the price discovery goes on, some users might want to withdraw their position - because they do not like how the price will be once the LP is launch. They can withdraw tokenA or tokenB.

price-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        let start_epoch = self.start_epoch().get();
+        let end_epoch = self.end_epoch().get();
+        require!(
+            start_epoch <= current_epoch,
+            "Deposit period not started yet"
+        );
+        require!(current_epoch < end_epoch, "Deposit period ended");
+
+        let launched_token_id = self.launched_token_id().get();
+        let current_launched_token_balance =
```
sasurobert, line 169: no need for this check as there is no need for the initial deposit.

price-discovery/src/redeem_token.rs
```diff
@@ -0,0 +1,133 @@
...
+
+pub const LAUNCHED_TOKEN_REDEEM_NONCE: u64 = 1;
+pub const ACCEPTED_TOKEN_REDEEM_NONCE: u64 = 2;
+
+#[elrond_wasm::module]
+pub trait RedeemTokenModule {
+    #[only_owner]
+    #[payable("EGLD")]
+    #[endpoint(issueRedeemToken)]
+    fn issue_redeem_token(
```
sasurobert, line 11: please integrate the new function which does everything in ONE go, issue esdt and local roles

price-discovery/src/create_pool.rs
dorin-iancu: This is basically just a callback for the dex `add_initial_liquidity`, so it can't be called directly. Additionally, it does not alter the state, so even if someone managed to somehow trick the DEX contracts to call this endpoint, it wouldn't affect anything.

price-discovery/src/lib.rs
dorin-iancu: Removed, now only the `deposit` endpoint exists.

price-discovery/src/lib.rs
dorin-iancu: Done.

price-discovery/src/lib.rs
dorin-iancu: Done.

price-discovery/src/lib.rs
dorin-iancu: Removed.

price-discovery/src/redeem_token.rs
dorin-iancu: Done.

price-discovery/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    fn compute_lp_amount_to_send(
+        &self,
+        redeem_token_nonce: u64,
+        redeem_token_amount: BigUint,
+    ) -> BigUint {
+        let total_lp_tokens = self.total_lp_tokens_received().get();
+
+        match redeem_token_nonce {
```
sasurobert, line 163: we need to make a common code with swap here - as this is critical to be the same. Right now it looks ok.

price-discovery/src/redeem_token.rs
```diff
@@ -0,0 +1,123 @@
+elrond_wasm::imports!();
+use hex_literal::hex;
+
+pub const LAUNCHED_TOKEN_REDEEM_NONCE: u64 = 1;
+pub const ACCEPTED_TOKEN_REDEEM_NONCE: u64 = 2;
+
+const SFT_TOKEN_TYPE_NAME: &[u8] = b"SFT";
```
sasurobert, line 7: META type has to be used. this is a financial token

dorin-iancu: Done.

price-discovery/src/lib.rs
dorin-iancu: Done. Now it's using the `quote` function from the `amm` module: https://github.com/ElrondNetwork/sc-dex-rs/blob/main/dex/pair/src/amm.rs#L16

sasurobert (approved):

price-discovery/src/lib.rs
dorin-iancu: Changed to use `rule_of_three` instead of `quote`

dex/pair/src/fee.rs
```diff
@@ -109,9 +109,10 @@ pub trait FeeModule:
     }
 
     fn reinject(&self, token: &TokenIdentifier, amount: &BigUint) {
-        let mut reserve = self.pair_reserve(token).get();
-        reserve += amount;
-        self.pair_reserve(token).set(&reserve);
+        self.pair_reserve(token)
+            .update(|reserve| *reserve += amount);
+        self.pair_virtual_reserve(token, token)
```
sasurobert, line 114: This is not needed - you need to redo the PR as you brought stuff from the main branch - which is not yet needed

dex/pair/src/liquidity_pool.rs
```diff
@@ -189,6 +225,16 @@ pub trait LiquidityPoolModule:
         self.calculate_k_constant(&first_token_amount, &second_token_amount)
     }
 
+    fn calculate_k_for_virtual_reserves(&self, token_side_id: &TokenIdentifier) -> BigUint {
```
sasurobert, line 228: these are functions from front-running-protection branch which was merged into main. We do not need these yet.

sasurobert (changes requested):

PR #242, opened 2022-01-17
Farm Staking SC

dorin-iancu:


dex/farm/src/lib.rs
```diff
@@ -229,11 +248,63 @@ pub trait Farm:
...
+        #[var_args] opt_accept_funds_func: OptionalArg<ManagedBuffer>,
+    ) -> SCResult<()> {
+        require!(self.is_active(), "Not active");
+        require!(!self.farm_token_id().is_empty(), "No farm token");
+
+        let farm_token_id = self.farm_token_id().get();
+        require!(payment_token_id == farm_token_id, "Bad input token");
+        require!(amount > 0, "Payment amount cannot be zero");
+
+        self.farming_token_total_liquidity()
```
sasurobert, line 272: this has to be moved to exitFarm -> unStakeFarm method. there are no rewards between unStake and unBond.

dex/farm/src/lib.rs
```diff
@@ -229,11 +248,63 @@ pub trait Farm:
...
+            &farm_token_id,
+            token_nonce,
+        );
+        let full_unlock_epoch = token_info
+            .decode_attributes::<UnbondSftAttributes>()?
+            .full_unlock_epoch;
+        let current_epoch = self.blockchain().get_block_epoch();
+
+        let farming_token_id = self.farming_token_id().get();
+        let farming_token_amount = if current_epoch < full_unlock_epoch {
```
sasurobert, line 286: let's remove penalty completely. this is not needed in staking contract

dex/farm/src/lib.rs
```diff
@@ -107,9 +112,12 @@ pub trait Farm:
         require!(token_in == farming_token_id, "Bad input token");
         require!(enter_amount > 0, "Cannot farm with amount of 0");
 
+        self.farming_token_total_liquidity()
```
sasurobert, line 115: why is this needed ? Where is this inside the farm contracts ?

sasurobert: oh, I see. For the bounded APR computation

dex/farm/src/lib.rs
dorin-iancu: Done. Now the unbond period is forced, with no early unbond allowed.

dex/farm/src/lib.rs
dorin-iancu: Done.

dex/farm/src/lib.rs
dorin-iancu: Yep. It's the only reason it's needed. Couldn't use SC balance directly in case of farming token being the same as the reward token.

dex/farm/src/custom_rewards.rs
```diff
@@ -1,54 +1,121 @@
...
+    ) -> BigUint {
+        let unbounded_rewards =
+            self.calculate_reward(amount, current_reward_per_share, initial_reward_per_share);
+        if unbounded_rewards == 0u32 {
+            return unbounded_rewards;
+        }
+
+        let farming_token_total_liquidity = self.farming_token_total_liquidity().get();
+        let max_apr = self.max_annual_percentage_rewards().get();
+        let max_rewards = farming_token_total_liquidity * max_apr
```
catalinnnn, line 50: not sure about this calculus :thinking: maybe we'll reconsider this together

dorin-iancu: Fixed in latest commit.

PR #245, opened 2022-01-18
Extend Vesting Precision

catalinnnn:


locked-asset/factory/src/locked_asset.rs
```diff
@@ -3,16 +3,25 @@ elrond_wasm::derive_imports!();
...
 
 #[derive(ManagedVecItem)]
 pub struct LockedToken<M: ManagedTypeApi> {
     pub token_amount: EsdtTokenPayment<M>,
-    pub attributes: LockedAssetTokenAttributes<M>,
+    pub attributes: LockedAssetTokenAttributesEx<M>,
+}
+
+#[derive(ManagedVecItem)]
+pub struct LockedTokenEx<M: ManagedTypeApi> {
```
sasurobert, line 22: I do not see any difference between LockedToken and LockedTokenEx as a structure. What am I missing ?

locked-asset/factory/src/lib.rs
```diff
@@ -51,9 +55,29 @@ pub trait LockedAssetFactory:
...
+    fn set_extended_attributes_activation_nonce(&self) {
+        //for extra safety
+        if self.extended_attributes_activation_nonce().is_empty() {
+            let one = BigUint::from(1u64);
+            let zero = BigUint::zero();
+            let mb_empty = ManagedBuffer::new();
+            let mv_empty = ManagedVec::new();
+            let token_id = self.locked_asset_token_id().get();
+
+            let nonce = self.send().esdt_nft_create(
```
sasurobert, line 71: in order to pass all the tests - add another check that the contract has already the nft create role

locked-asset/factory/src/locked_asset.rs
catalinnnn: LockedAssetTokenAttributes```Ex```

locked-asset/factory/src/lib.rs
catalinnnn: added a check for init_epoch not to be empty

locked-asset/factory/src/locked_asset.rs
sasurobert: Both are with "Ex"

catalinnnn: ohhh, right, should have been without Ex, the old struct I mean. removed the struct entirely since only the Ex version will be used from now on

sasurobert (dismissed):

locked-asset/factory/src/lib.rs
yzhang90: This `esdt_nft_create` is to get the latest nonce?

catalinnnn: yes

locked-asset/factory/src/locked_asset.rs
```diff
@@ -94,22 +96,23 @@ pub trait LockedAssetModule: token_send::TokenSendModule {
...
         let mut unlock_milestones_merged =
-            ArrayVec::<UnlockMilestoneExtended, MAX_MILESTONES_IN_SCHEDULE>::new();
+            ArrayVec::<UnlockMilestoneEx, MAX_MILESTONES_IN_SCHEDULE>::new();
         for old_milestone in old_unlock_milestones.iter() {
             if old_milestone.unlock_epoch > current_epoch {
-                let new_unlock_percent: u64 =
-                    (old_milestone.unlock_percent as u64) * ONE_MILLION / unlock_percent_remaining;
-                unlock_milestones_merged.push(UnlockMilestoneExtended {
+                let new_unlock_percent =
+                    old_milestone.unlock_percent * PRECISION_EX_INCREASE * ONE_MILLION
```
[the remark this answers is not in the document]
catalinnnn: unlock_percent / unlock_percent_remeining gives you the fraction of what unlock_percent is in comparison with unlock_percent_remaining. in order to bump this to a percent format (based on the new precision), we need to multiply with PRECISION_EX_INCREASE

claudiulataretu (approved):

PR #251, opened 2022-01-26
Migration to v1.4

catalinnnn:


dex/farm/src/lib.rs
```diff
@@ -613,4 +613,207 @@ pub trait Farm:
...
+
+        self.end_produce_rewards();
+        self.reward_per_share().get()
+    }
+
+    // In the new sc, we have to set the rps, so rewards can continue
+    // with positions being untouched.
+    #[only_owner]
+    #[endpoint(setRpsAndStartRewards)]
+    fn set_rps_and_start_rewards(&self, rps: BigUint) {
```
sasurobert, line 796: This is a nice IDEA! It is somewhat dangerous. My first thought was that we make the transition process as normal and users will not get rewards on the old contract - only from the new - and the new contract will start accumulating position as normal. But this is much beneficial for the user - as it would get the rewards, and there is no need to rush for the migration.

locked-asset/proxy-dex/src/proxy_farm.rs
```diff
@@ -370,6 +370,73 @@ pub trait ProxyFarmModule:
...
+        let new_pos = self
+            .farm_contract_proxy(farm_address.clone())
+            .migrate_to_new_farm(OptionalArg::Some(self.blockchain().get_sc_address()))
+            .add_token_transfer(farm_token_id, farm_token_nonce, farm_amount)
+            .execute_on_dest_context_custom_range(|_, after| (after - 1, after));
+
+        // Burn the old proxy farm position
+        self.send().esdt_local_burn(&token_id, token_nonce, &amount);
+
+        // Create a new proxy farm position based on the new farm position.
```
sasurobert, line 418: Nice and clean!

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,207 @@ pub trait Farm:
...
+            old_farm_address,
+            old_farm_token_id,
+            new_farm_address,
+        });
+    }
+
+    // We also need to get the rps and transfer it to the new SC.
+    #[only_owner]
+    #[endpoint(stopRewardsAndGetRps)]
+    fn stop_rewards_and_get_rps(&self) -> BigUint {
```
sasurobert, line 783: I think it would be better to directly have a dunction of migrate rewards and RPS. So there would be no need to 2 transactions from the owner. Only one. And contract calling another contract setting up the values.

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,207 @@ pub trait Farm:
             &context.get_input_attributes().unwrap().compounded_reward,
         )
     }
+
+    //This function is neccessary in old farms.
+    #[payable("*")]
+    #[endpoint(migrateToNewFarm)]
+    fn migrate_to_new_farm(
```
sasurobert, line 620: Let's disable all endpoints when migration starts. ExitFarm and migrateToNewFarm should be enabled only.

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,207 @@ pub trait Farm:
...
+            self.generate_aggregated_rewards(context.get_storage_cache_mut());
+            self.calculate_reward(&mut context);
+            context.decrease_reward_reserve();
+            self.calculate_initial_farming_amount(&mut context);
+            self.increase_reward_with_compounded_rewards(&mut context);
+            self.commit_changes(&context);
+
+            // Update the farm supply even though we dont burn the tokens by ourselves.
+            // The position will be burned and this supply has to be updated.
+            self.farm_token_supply()
```
sasurobert, line 668: on exitFarm we takeout and send the farming tokens to the new contract - that is the actual liquidity which has to be sent back to the user when he does exitFarm from the new contract.

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,207 @@ pub trait Farm:
...
+                &self.blockchain().get_sc_address(),
+                &old_position.token_identifier,
+                old_position.token_nonce,
+            )
+            .decode_attributes()
+            .unwrap();
+
+        // Do not call burn_farm_tokens since this farm tokens belong to other contract
+        // which already updated its farm token supply counter.
+        self.send().esdt_local_burn(
```
sasurobert, line 733: we have to give burn roles to this contract. which is not implemented in the old farm.

sasurobert (changes requested):

claudiu725 (changes requested):
Clippy complains:
an unused `elrond_wasm::types::SCResult` that must be used
in ` dex/farm/src/lib.rs`
This means that a check somewhere gets ignored!

dex/farm/src/lib.rs
sasurobert: Done. I have written it.

dex/farm/src/lib.rs
sasurobert: done

dex/farm/src/lib.rs
sasurobert: We cannot go with this code, as we will move from contract v1.2 to v1.4 split farms where user decides in which farm he wants to go - the rewards per share cannot be used. We need to make enterFarmWithOriginalEpoch.

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,214 @@ pub trait Farm:
...
+            );
+
+            self.load_farm_token_id(&mut context);
+            assert!(
+                self,
+                !context.get_farm_token_id().unwrap().is_empty(),
+                ERROR_NO_FARM_TOKEN,
+            );
+
+            self.load_farming_token_id(&mut context);
```
claudiulataretu, line 647: might need a check that farming_token_id is not empty? It is the same case as with farm_token_id

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,214 @@ pub trait Farm:
...
+
+            // Update the farm supply even though we dont burn the tokens by ourselves.
+            // The position will be burned and this supply has to be updated.
+            self.farm_token_supply()
+                .update(|x| *x -= &context.get_tx_input().get_payments().get_first().amount);
+        }
+
+        let mut payments = ManagedVec::<Self::Api, EsdtTokenPayment<Self::Api>>::new();
+        payments.push(context.get_tx_input().get_payments().get_first().clone());
+        payments.push(context.get_final_reward().unwrap().clone());
```
claudiulataretu, line 674: When you are migrating from old farm to new farm it is supposed to also send the actual farming tokens. Does these two payments includes also the actual farming tokens?

catalinnnn: they do not. as stated, they contain the farm token and the reward token

dex/farm/src/lib.rs
catalinnnn: if it is empty it will signal error so its fine without the check

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,214 @@ pub trait Farm:
...
+    #[endpoint(stopRewardsAndMigrateRps)]
+    fn stop_rewards_and_migrate_rps(&self) {
+        assert!(self, !self.farm_configuration().is_empty(), b"empty config");
+        let config = self.farm_configuration().get();
+        assert!(self, config.is_old, b"bad config");
+
+        self.state().set(&State::Migrate);
+        self.end_produce_rewards();
+
+        self.self_proxy(config.new_farm_address)
```
sasurobert, line 792: Do we migrate the same RPS for both of the contracts ?

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,214 @@ pub trait Farm:
...
+            is_old,
+            old_farm_address,
+            old_farm_token_id,
+            new_farm_address,
+        });
+    }
+
+    // We also need to get the rps and transfer it to the new SC.
+    #[only_owner]
+    #[endpoint(stopRewardsAndMigrateRps)]
```
sasurobert, line 783: This endpoint must be moved to v1_2 version of contracts. As those has to be updated first. We need to add test there.

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,214 @@ pub trait Farm:
             &context.get_input_attributes().unwrap().compounded_reward,
         )
     }
+
+    //This function is neccessary in old farms.
+    #[payable("*")]
+    #[endpoint(migrateToNewFarm)]
```
sasurobert, line 619: This endpoint must be moved to v1_2 version of contracts. As those has to be updated first. We need to add test there.

dex/farm/src/lib.rs
```diff
@@ -613,4 +613,214 @@ pub trait Farm:
...
+            // The position will be burned and this supply has to be updated.
+            self.farm_token_supply()
+                .update(|x| *x -= &context.get_tx_input().get_payments().get_first().amount);
+        }
+
+        let mut payments = ManagedVec::<Self::Api, EsdtTokenPayment<Self::Api>>::new();
+        payments.push(context.get_tx_input().get_payments().get_first().clone());
+        payments.push(context.get_final_reward().unwrap().clone());
+
+        self.self_proxy(new_farm_address)
```
sasurobert, line 676: we need 2 addresses here - and choose address 1 or address 2 depending whether the user is farming locked mex or not.

dex/farm/src/lib.rs
sasurobert: We need 2 addresses and to call both contracts.

sasurobert (changes requested):

dex/farm/src/migration_from_v1_2.rs
```diff
@@ -0,0 +1,191 @@
...
+    + farm_token::FarmTokenModule
+    + rewards::RewardsModule
+{
+    #[payable("*")]
+    #[endpoint(migrateFromV1_2Farm)]
+    fn migrate_from_v1_2_farm(
+        &self,
+        #[var_args] orig_caller_opt: OptionalArg<ManagedAddress>,
+    ) -> EsdtTokenPayment<Self::Api> {
+        require!(self.farm_migration_config().is_empty(), "empty config");
```
sasurobert, line 20: add check for state to be active.

dex/farm_with_lock/src/migration_from_v1_2.rs
```diff
@@ -0,0 +1,191 @@
...
+#[elrond_wasm::module]
+pub trait MigrationModule:
+    config::ConfigModule
+    + token_send::TokenSendModule
+    + farm_token::FarmTokenModule
+    + rewards::RewardsModule
+{
+    #[payable("*")]
+    #[endpoint(migrateFromV1_2Farm)]
+    fn migrate_from_v1_2_farm(
```
sasurobert, line 16: try to move as much code as possible together into one common file.

locked-asset/proxy-dex/src/migration_from_v1_2.rs
```diff
@@ -0,0 +1,107 @@
...
+        // The actual work starts here
+        let wrapped_farm_token_attrs =
+            self.get_wrapped_farm_token_attributes(&payment_token_id, payment_token_nonce);
+        let farm_token_id = wrapped_farm_token_attrs.farm_token_id.clone();
+        let farm_token_nonce = wrapped_farm_token_attrs.farm_token_nonce;
+        let farm_amount = payment_amount.clone();
+
+        // Get the new farm position from the new contract.
+        let new_pos = self
+            .farm_v1_2_contract_proxy(farm_address.clone())
```
sasurobert, line 75: add clippy fixes. delete redundant clone.

sasurobert (commented):
Logic looks good! Super nice! Need some test :)

sasurobert (dismissed):

claudiu725 (approved):

common/modules/farm/migration_from_v1_2/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+{
+    #[payable("*")]
+    #[endpoint(migrateFromV1_2Farm)]
+    fn migrate_from_v1_2_farm(
+        &self,
+        #[var_args] orig_caller_opt: OptionalArg<ManagedAddress>,
+    ) -> EsdtTokenPayment<Self::Api> {
+        require!(self.state().get() == State::Active, "not active");
+
+        require!(self.farm_migration_config().is_empty(), "empty config");
```
virgil-serbanuta, line 37: Should this condition be negated?

catalinnnn: done

sasurobert (approved):

PR #261, opened 2022-02-01
Minimize number of Milestones upon merge

catalinnnn:


alyn509 (approved):

alyn509 (dismissed):

claudiu725 (approved):

locked-asset/factory/src/locked_asset_token_merge.rs
```diff
@@ -163,9 +163,9 @@ pub trait LockedAssetTokenMergeModule:
         for elem in array.iter() {
             let last = unlock_epoch_amount_merged.last().unwrap_or(&default);
 
-            if elem.epoch == last.epoch {
+            if elem.epoch == last.epoch || elem.epoch - 1 == last.epoch {
```
dorin-iancu, line 166: nitpicking, but I think it's more readable if you rewrite it as
```suggestion
            if elem.epoch == last.epoch || elem.epoch == last.epoch + 1 {
```

Since that's what you want to check, after all. If the element's epoch is equal to the last or last + 1.

dorin-iancu (dismissed):

alyn509 (approved):

sasurobert (approved):

PR #265, opened 2022-02-03
Mandatory params v1.2

catalinnnn:


dex/farm/src/migration.rs
```diff
@@ -96,7 +96,7 @@ pub trait MigrationModule:
 
         Ok(self
             .farm_v1_4_contract_proxy(new_farm_dest)
-            .migrate_from_v1_2_farm(orig_caller_opt)
+            .migrate_from_v1_2_farm(orig_caller)
```
camilbancioiu, line 99: Wouldn't `self.blockchain().get_caller()` work here? It doesn't need to be an argument to the endpoint.

catalinnnn: User calls Old Farm that calls New Farm. New Farm should send the tokens to User

catalinnnn: Also, User can Call Proxy that Calls Old Farm that calls New Farm. New Farm should send the tokens to User

catalinnnn: So an Orig Caller is needed

camilbancioiu (approved):

sasurobert (approved):

PR #268, opened 2022-02-03
Permission fixes

catalinnnn:
Added #[only_owner]
SetRpsAndStartRewards fix Perms
Removed redundant check

### DIFF
--- a/dex/farm-staking-proxy/wasm/src/lib.rs
+++ b/dex/farm-staking-proxy/wasm/src/lib.rs
@@ -7,7 +7,6 @@
 elrond_wasm_node::wasm_endpoints! {
     farm_staking_proxy
     (
-        init
         callBack
         claimDualYield
         getDualYieldTokenId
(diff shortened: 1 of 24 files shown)


camilbancioiu (approved):

sasurobert (approved):

dex/farm-staking-proxy/wasm/src/lib.rs
```diff
@@ -7,7 +7,6 @@
 elrond_wasm_node::wasm_endpoints! {
     farm_staking_proxy
     (
-        init
```
sasurobert, line 10: no init ?

catalinnnn: yep... strange, the init is still there, compiled it manually on my machine, but this auto generated code does not include it anymore for some reason, probably the rust framework minor version changed and it just doesnt generate init anymore

PR #269, opened 2022-02-03
Permissions V1.2

catalinnnn:
Fix perms on stopRewardsAndGetRps
Removed redundant branches

### DIFF
--- a/dex/farm/src/rewards.rs
+++ b/dex/farm/src/rewards.rs
@@ -152,8 +152,12 @@ pub trait RewardsModule:
     }
 
     #[endpoint]
-    fn end_produce_rewards(&self) -> SCResult<()> {
+    fn end_produce_rewards_as_owner(&self) -> SCResult<()> {
         self.require_permissions()?;
+        self.end_produce_rewards()
+    }
+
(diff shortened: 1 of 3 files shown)


dex/farm/src/rewards.rs
```diff
@@ -152,8 +152,12 @@ pub trait RewardsModule:
     }
 
     #[endpoint]
-    fn end_produce_rewards(&self) -> SCResult<()> {
+    fn end_produce_rewards_as_owner(&self) -> SCResult<()> {
```
camilbancioiu, line 155: `#[only_owner]` here as well? :)

sasurobert: no need - as at the next line you have self.require_permissions. code is good.

sasurobert (approved):

sasurobert (approved):

dex/farm/src/rewards.rs
catalinnnn: I prefer either all only_owner or all require perms and I wouldnt change all to only_owner since its the v2 SCs and will soon be unused

camilbancioiu (approved):

PR #270, opened 2022-02-03
Fix price discovery setup

claudiulataretu:


dex/pair/src/lib.rs
```diff
@@ -146,6 +146,8 @@ pub trait Pair<ContractReader>:
         self.execute_output_payments(&context);
         self.emit_add_liquidity_event(&context);
 
+        self.initial_liquidity_adder().clear();
```
sasurobert, line 149: why do you need this? you just want to clear the storage ?

price-discovery/src/lib.rs
```diff
@@ -50,7 +45,6 @@ pub trait PriceDiscovery:
             "Start epoch must be before end epoch"
         );
 
-        self.dex_sc_address().set(&dex_sc_address);
         self.launched_token_id().set(&launched_token_id);
```
sasurobert, line 48: do we already have an only owner function which can set the dex sc address ?

dex/pair/src/lib.rs
sasurobert: I see - this needs to be empty on normal addLiquidity

price-discovery/src/lib.rs
claudiulataretu: I added an only_owner function for this set with this PR

dex/pair/src/lib.rs
claudiulataretu: There is a check in the normal `addLiquidity` endpoint to not have a price discovery address set in order to add liquidity

catalinnnn (approved):

PR #291, opened 2022-02-11
Bot protection

catalinnnn:


dex/pair/src/bot_protection.rs
```diff
@@ -0,0 +1,196 @@
...
+            return;
+        }
+
+        let current_liquidity = ctx.get_lp_token_supply();
+        if current_liquidity == &0u64 {
+            return;
+        }
+
+        let caller = ctx.get_caller();
+        let num_adds = self.num_adds_by_address(caller).get();
```
sasurobert, line 112: nice addition - but we need to clean the addresses after :). somehow.

dex/pair/src/bot_protection.rs
```diff
@@ -0,0 +1,196 @@
...
+        let amount_out_percent = final_out * PERCENT_MAX / reserve_out;
+        require!(
+            amount_out_percent < bp_config.volume_percent,
+            "swap amount out too large"
+        );
+
+        self.num_swaps_by_address(caller).set(&(num_swaps + 1));
+    }
+
+    fn require_can_proceed_remove(&self, ctx: &RemoveLiquidityContext<Self::Api>) {
```
sasurobert, line 62: why do we need protection for remove and add ?

dex/pair/src/bot_protection.rs
catalinnnn: done

dex/pair/src/bot_protection.rs
catalinnnn: because swap works with volume percents, one whale can manipulate the volumes and make high volume swaps by after adding much volume. this protection takes care of that by not allowing anyone to high add or remove (thus manipulating allowed the volume amount on swap - being fixed percent of total)

sasurobert (approved):

PR #292, opened 2022-02-11
Price discovery restrictions

dorin-iancu:
There are now multiple phases during the price discovery period:
- phase 1: Anyone can deposit and withdraw any amounts
- phase 2: Anyone can deposit, but withdrawals come with a linear increasing penalty
- phase 3: Deposits are not allowed. Withdrawals are allowed with a fixed penalty
- phase 4: Adding liquidity to the LP and allowing users to redeem after a certain number of epochs

Accumulated penalties are given as rewards in the redeem period, according to the percentage of the total pool the user owns. (i.e. if the user owns 50% of the initial liquidity, they will receive 50% of the accumulated penalties for that specific token)

### DIFF
--- a/price-discovery/src/common_storage.rs
+++ b/price-discovery/src/common_storage.rs
@@ -26,11 +26,11 @@ pub trait CommonStorageModule {
     #[storage_mapper("totalLpTokensReceived")]
     fn total_lp_tokens_received(&self) -> SingleValueMapper<BigUint>;
 
-    #[view(getStartEpoch)]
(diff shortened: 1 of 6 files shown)


sasurobert (dismissed):
It would be nice to offer some rewards as well for those who lock their LPs. Add a depositRewards endpoint which can be called by a whitelisted address and with a given token, until unlock period starts. And when users are unlocking they should get their fair share of rewards.

catalinnnn (approved):

sasurobert (approved):

PR #315, opened 2022-03-02
Governance

catalinnnn:


psorinionut (approved):

dex/governance/src/contract.rs
```diff
@@ -0,0 +1,138 @@
...
+        let proposal = self.proposal(attr.proposal_id).get();
+
+        let pstat = self.get_proposal_status(&proposal);
+        match pstat {
+            ProposalStatus::Defeated | ProposalStatus::Executed => {
+                self.send_back(attr.payment);
+                self.burn_vote_nft(payment);
+            }
+            ProposalStatus::Succeeded => {
+                sc_panic!(PROPOSAL_NEEDS_TO_BE_EXECUTED)
```
sasurobert, line 131: I think tokens can be redeemed back if it was  already voted - no need for execution. But I see your point - however, people could block tokens in some proposals which will never be executed as someone veto's is.

PR #322, opened 2022-03-09
Price discovery improvements

dorin-iancu:
- owner can now select a min-price for the launched token. The check is only activated after the initial launched tokens are deposited
- add an option for owner to add extra rewards. Users will receive their share based on the ratio of lp tokens the are entitled to.
- fix extra lp tokens calculation to also use the ratio of lp tokens from total

### DIFF
--- a/price-discovery/src/lib.rs
+++ b/price-discovery/src/lib.rs
@@ -170,10 +170,6 @@ pub trait PriceDiscovery:
 
         let penalty_percentage = phase.to_penalty_percentage();
         let penalty_amount = &payment_amount * &penalty_percentage / MAX_PERCENTAGE;
-        if penalty_amount > 0 {
-            self.accumulated_penalty(payment_nonce)
-                .update(|p| *p += &penalty_amount);
-        }
 
         let caller = self.blockchain().get_caller();
         let withdraw_amount = payment_amount - penalty_amount;
@@ -197,7 +193,7 @@ pub trait PriceDiscovery:
         require!(payment_token == redeem_token_id, INVALID_PAYMENT_ERR_MSG);
 
         let rewards = self.compute_rewards(payment_nonce, &payment_amount);
-        self.burn_redeem_token(payment_nonce, &payment_amount);
+        self.burn_redeem_token_without_supply_decrease(payment_nonce, &payment_amount);
 
         let caller = self.blockchain().get_caller();
         if rewards.lp_tokens_amount > 0 {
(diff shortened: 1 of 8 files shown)


price-discovery/src/lib.rs
```diff
@@ -243,4 +290,27 @@ pub trait PriceDiscovery:
...
+        let launched_token_balance = self.blockchain().get_sc_balance(&launched_token_id, 0);
+        let accepted_token_balance = self.blockchain().get_sc_balance(&accepted_token_id, 0);
+
+        if accepted_token_balance == 0 {
+            return;
+        }
+
+        require!(launched_token_balance > 0, "No launched tokens available");
+        require!(
+            accepted_token_balance * MIN_PRICE_PRECISION / launched_token_balance >= min_price,
```
claudiulataretu, line 308: maybe extract this price compute in a variable

claudiu725 (dismissed):

price-discovery/src/lib.rs
claudiulataretu: I think `MIN_PRICE_PRECISION` should be set based on `launched_token` decimals numbers

sasurobert: This can be changed on deploy - I do not see it necessary to be in config or kept in storage

sasurobert (approved):
Looks good!

price-discovery/src/lib.rs
```diff
@@ -185,49 +225,56 @@ pub trait PriceDiscovery:
...
-            self.get_percentage_of_total_supply(redeem_token_nonce, redeem_token_amount);
-
-        let extra_lp_tokens = self.extra_lp_tokens().get();
-        let bonus = percentage_of_redeeem_token_supply * extra_lp_tokens / MAX_PERCENTAGE / 2u32;
-        self.extra_lp_tokens()
-            .update(|extra_amt| *extra_amt -= &bonus);
-
-        match redeem_token_nonce {
+    ) -> RewardsPair<Self::Api> {
+        let total_lp_tokens = self.total_claimable_lp_tokens().get();
```
sasurobert, line 256: total_lp_received would be better here.

price-discovery/src/lib.rs
```diff
@@ -185,49 +225,56 @@ pub trait PriceDiscovery:
...
-                let base_lp_amount =
-                    redeem_token_amount * &total_lp_tokens / accepted_token_final_amount / 2u32;
-
-                base_lp_amount + bonus
+                redeem_token_amount * &total_lp_tokens / accepted_token_final_amount / 2u32
             }
             _ => BigUint::zero(),
+        };
+
+        let total_extra_lp_tokens = self.extra_lp_tokens().get();
```
sasurobert, line 269: there is no extra lp. as everything was put into swap SC

sasurobert (changes requested):
for above comments

claudiu725 (approved):

sasurobert (approved):

PR #342, opened 2022-03-18
Price Discovery Events

dorin-iancu:
Events for:
- deposit extra rewards
- deposit
- withdraw
- create pool
- redeem


dex/price-discovery/src/events.rs
```diff
@@ -0,0 +1,60 @@
+elrond_wasm::imports!();
+
+#[elrond_wasm::module]
+pub trait EventsModule {
+    #[event("depositExtraRewardsEvent")]
+    fn deposit_extra_rewards_event(
```
claudiulataretu, line 6: might be a good idea to also add the token_id for the rewards token

dex/price-discovery/src/events.rs
```diff
@@ -0,0 +1,60 @@
...
+        &self,
+        #[indexed] block: u64,
+        #[indexed] caller: &ManagedAddress,
+        amount: &BigUint,
+    );
+
+    #[event("depositEvent")]
+    fn deposit_event(
```
claudiulataretu, line 14: for `deposit` and `withdraw` events would be nice to have also the amounts for launched tokens and accepted tokens

dorin-iancu: Done.

claudiulataretu (approved):

PR #345, opened 2022-03-21
Check token ids

dorin-iancu:
- check that launched token, accepted token and extra rewards token are all different


claudiu725 (approved):

dorin-iancu:
Will have to rewrite.

PR #346, opened 2022-03-21
balance in storage

dorin-iancu:
Keep balance in storage instead of relying on get_sc_balance. This matters when reward token is the same as accepted token or launched token.


claudiu725 (approved):

sasurobert (commented):
some unit test here - for the new functionality is needed.

PR #349, opened 2022-03-22
soften min price require

dorin-iancu:
- Accepted token deposits are now allowed if the price is below min_price before and after the deposit. This helps when a lot of Launched tokens are deposited at the start, otherwise, one user would have to deposit a lot of Accepted tokens right after

### DIFF
--- a/dex/price-discovery/src/lib.rs
+++ b/dex/price-discovery/src/lib.rs
@@ -14,6 +14,7 @@ pub mod phase;
 pub mod redeem_token;
 
 const INVALID_PAYMENT_ERR_MSG: &[u8] = b"Invalid payment token";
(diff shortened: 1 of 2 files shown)


dex/price-discovery/src/lib.rs
```diff
@@ -342,21 +355,56 @@ pub trait PriceDiscovery:
...
+        }
+
+        require!(
+            launched_token_balance_after > 0,
+            "No launched tokens available"
+        );
+
+        let price_before = self.calculate_price(
+            &accepted_token_balance_before,
+            &launched_token_balance_before,
```
psorinionut: Clippy warnings

psorinionut (approved):

claudiu725 (approved):

sasurobert (approved):

PR #351, opened 2022-03-23
claim dual yield single token only

dorin-iancu:
- `claim_dual_yield` from `farm_staking_proxy` now only works with a single payment, and furthermore, `claim_rewards_with_new_value` from `farm_staking` was also cleanup to only work with one payment and one new_value.

Since most of the code was merging code, `claim_rewards_common` was deleted, and both claim functions from `farm_staking` have their own implementation

### DIFF
--- a/farm-staking/farm-staking/src/lib.rs
+++ b/farm-staking/farm-staking/src/lib.rs
@@ -306,47 +306,90 @@ pub trait Farm:
     fn claim_rewards(&self) -> ClaimRewardsResultType<Self::Api> {
         let payments = self.call_value().all_esdt_transfers();
 
-        self.claim_rewards_common(payments, None)
+        require!(self.is_active(), "Not active");
+        require!(!self.farm_token_id().is_empty(), "No farm token");
+
+        let first_payment = payments
+            .try_get(0)
+            .unwrap_or_else(|| sc_panic!("empty payments"));
+        let additional_payments = payments.slice(1, payments.len()).unwrap_or_default();
+
+        let payment_token_id = &first_payment.token_identifier;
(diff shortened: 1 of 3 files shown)


sasurobert (approved):

farm-staking/farm-staking/src/lib.rs
```diff
@@ -306,47 +306,90 @@ pub trait Farm:
...
+            old_farming_amount,
+            &current_reward_per_share,
+            &farm_attributes.reward_per_share,
+        );
+        let new_compound_reward_amount = self.rule_of_three(
+            old_farming_amount,
+            &farm_attributes.current_farm_amount,
+            &farm_attributes.compounded_reward,
         );
+        let new_farming_amount = old_farming_amount;
```
claudiulataretu, line 341: Why are you assigning the `old_farming_amount` to `new_farming_amount` and not use a single variable `farming_amount`? Should be the same thing

farm-staking/farm-staking/src/lib.rs
```diff
@@ -306,47 +306,90 @@ pub trait Farm:
...
+        let reward_token_id = self.reward_token_id().get();
+        self.generate_aggregated_rewards();
+
+        let current_reward_per_share = self.reward_per_share().get();
+        let reward = self.calculate_reward(
+            old_farming_amount,
+            &current_reward_per_share,
+            &farm_attributes.reward_per_share,
+        );
+        let new_compound_reward_amount = self.rule_of_three(
```
claudiulataretu, line 336: compounded reward should not change by the claim rewards. I don't see a reason to compute it again.

claudiulataretu (approved):

PR #356, opened 2022-03-24
Use elrond-actions contracts.yml v1

claudiu725:
Use [`contracts.yml`](https://github.com/ElrondNetwork/elrond-actions/blob/v1/.github/workflows/contracts.yml). The rust version is the same as in the release workflow. The vmtools version was upgraded.
Side notes:
- `./build-wasm.sh` and `./wasm-opt.sh` are not used for github actions anymore since `erdpy contract build -r` is used now

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -9,8 +9,8 @@ jobs:
     name: Contracts
     uses: ElrondNetwork/elrond-actions/.github/workflows/contracts.yml@v1
     with:
-      rust-toolchain: nightly-2021-11-04
-      vmtools-version: v1.4.32
+      rust-toolchain: nightly-2022-03-01
+      vmtools-version: v1.4.43
       extra-build-args: --ignore-eei-checks
     secrets:
       token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -1,109 +1,16 @@
...
-            default: true
-      - uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-
+  contracts:
+    name: Contracts
+    uses: ElrondNetwork/elrond-actions/.github/workflows/contracts.yml@v1
+    with:
+      rust-toolchain: nightly-2021-11-04
```
andreibancioiu, line 12: The release workflow uses the docker image `elrondnetwork/elrond-sdk-erdpy-rust:frozen-003`, which has:

```
LABEL erdpy=1.2.2
LABEL rust=nightly-2022-03-01
LABEL wasm-opt=v1.3.0
LABEL wasm-opt-binaryen=version_105
LABEL nodejs=v12.18.3
```

Therefore, perhaps use rust `nightly-2022-03-01` (for consistency)?

andreibancioiu (approved):

PR #359, opened 2022-03-25
renames + cleanup

dorin-iancu:
- renamed rewards to "bought tokens"
- removed commented code from tests
- removed unused structs/events
- fixed redeem event

### DIFF
--- a/dex/price-discovery/src/lib.rs
+++ b/dex/price-discovery/src/lib.rs
@@ -206,9 +206,10 @@ pub trait PriceDiscovery:
         );
     }
 
-    /// After the liquidity pool has been created and the LP tokens received,
-    /// users can withdraw their fair share of the LP tokens by depositing their SFTs
-    /// and a share of the extra rewards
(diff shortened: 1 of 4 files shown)


dex/price-discovery/src/lib.rs
```diff
@@ -206,9 +206,10 @@ pub trait PriceDiscovery:
         );
     }
 
-    /// After the liquidity pool has been created and the LP tokens received,
-    /// users can withdraw their fair share of the LP tokens by depositing their SFTs
-    /// and a share of the extra rewards
+    /// After the unbond period has ended,
```
sasurobert, line 209: this has to be changed. as not after unbound - but after initial price discovery phase

sasurobert (approved):

dex/price-discovery/src/lib.rs
dorin-iancu: In the next PR :)
The current description was made to match the current implementation

PR #365, opened 2022-03-28
simple lock in pair

dorin-iancu:
- extracted the `locking_module` from price discovery into the common modules folder, and extended it with the "lock" function (without forwarding) and `set_unlock_epoch` endpoint
- used `simple-lock` in the `pair` contract instead of the `locked-asset-factory`
- updated & fixed tests


psorinionut (approved):

dex/pair/src/ctx_helper.rs
```diff
@@ -201,6 +201,13 @@ pub trait CtxHelper:
         context.set_output_payments(payments);
     }
 
+    fn should_generate_locked_asset(&self) -> bool {
+        let current_epoch = self.blockchain().get_block_epoch();
+        let unlock_epoch = self.unlock_epoch().get();
```
sasurobert, line 206: we will LOCK in the first epoch only - but UNLOCK happens in like epoch 3. so we need another storage map

dorin-iancu: Done. Also moved into a separate `locking_wrapper` module.

sasurobert (approved):

PR #370, opened 2022-03-30
enforce token order in lp

dorin-iancu:
- change whitelist to a mapping of (token A, token B) -> LP address
- add checks for received payments


catalinnnn (approved):

sasurobert (commented):
Green!

PR #372, opened 2022-03-30
Token attributes to nonce mapping

dorin-iancu:
- (token_id, attributes) pairs are now mapped to specific token nonces, so tokens that have the exact same attributes no longer create different nonces for each locking/adding to lp


catalinnnn (approved):

sasurobert (approved):
Nice!

PR #373, opened 2022-03-31
farm integration

dorin-iancu:
- integrate enter, exit, claim rewards, compound rewards with locked tokens

### DIFF
--- a/locked-asset/simple-lock/src/proxy_farm.rs
+++ b/locked-asset/simple-lock/src/proxy_farm.rs
@@ -264,7 +264,7 @@ pub trait ProxyFarmModule:
         farming_token_id: &TokenIdentifier,
         farm_type: FarmType,
     ) -> ManagedAddress {
-        let mapper = self.farm_address_for_token(&farming_token_id, farm_type);
+        let mapper = self.farm_address_for_token(farming_token_id, farm_type);
         require!(
             !mapper.is_empty(),
             "No farm address for the specified token and type pair",
--- a/locked-asset/simple-lock/src/proxy_farm.rs
+++ b/locked-asset/simple-lock/src/proxy_farm.rs
@@ -1,0 +1,303 @@
+elrond_wasm::imports!();
+elrond_wasm::derive_imports!();
+
+use crate::{error_messages::*, locked_token::LockedTokenAttributes};
+
+#[derive(
+    TypeAbi, TopEncode, TopDecode, NestedEncode, NestedDecode, PartialEq, Debug, Clone, Copy,
+)]
+pub enum FarmType {
+    SimpleFarm,
+    FarmWithLockedRewards,
+}
+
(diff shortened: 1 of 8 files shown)


locked-asset/simple-lock/src/proxy_farm.rs
```diff
@@ -0,0 +1,303 @@
...
+
+#[elrond_wasm::module]
+pub trait ProxyFarmModule:
+    crate::farm_interactions::FarmInteractionsModule
+    + crate::locked_token::LockedTokenModule
+    + elrond_wasm_modules::default_issue_callbacks::DefaultIssueCallbacksModule
+{
+    #[only_owner]
+    #[payable("EGLD")]
+    #[endpoint(issueFarmProxyToken)]
```
sasurobert, line 37: I think we can have another strategy for these kinds of setups - as we do not necessary need to do this from a contract. We could issue the token, setroles for the contract and give the ownership to the contract. so we will not need to do these proxy stuff - in a set of contract this might work @catalinnnn @claudiu725 ?

locked-asset/simple-lock/src/proxy_farm.rs
```diff
@@ -0,0 +1,303 @@
...
+            &farm_proxy_token_attributes,
+        )
+    }
+
+    fn try_get_farm_address(
+        &self,
+        farming_token_id: &TokenIdentifier,
+        farm_type: FarmType,
+    ) -> ManagedAddress {
+        let mapper = self.farm_address_for_token(&farming_token_id, farm_type);
```
sasurobert, line 267: clippy

locked-asset/simple-lock/src/proxy_farm.rs
dorin-iancu: It could work, but I think the current implementation needs fewer setup steps, and also ensures that everything worked as intended (as long as the transactions went through successfully)

In the current version, we only need two transactions: issue and set roles.
In the version you're suggesting, we would need more steps:
- issue the token
- get the token ID from the issue transaction
- set the local roles on the contract's address
- change token ownership to the contract
- set the token ID in the contract

locked-asset/simple-lock/src/proxy_farm.rs
dorin-iancu: Fixed

locked-asset/simple-lock/src/proxy_farm.rs
catalinnnn: I think manual issue and setting roles are the way to go, especially if we have the ability to transfer the token ownership to the contract, which I didn't know about

sasurobert: You can transfer ownership of tokenID to a contract.

sasurobert (approved):

PR #384, opened 2022-04-06
fix issue

dorin-iancu:


sasurobert (approved):
Reverted to old code :D

sasurobert (commented):
I think we should use the new endpoint which gives roles and everything to the metaESDT. And a simple separate call which does the first NFT creates. The current code has 2 callbacks and 2 asyncCalls. The other version has only 1 of those.

PR #421, opened 2022-07-13
governance with energy

dorin-iancu:


energy-integration/governance-v2/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+    + proposal_storage::ProposalStorageModule
+    + views::ViewsModule
+    + energy_query_module::EnergyQueryModule
+{
+    // Used to deposit tokens for "payable" actions
+    // There is no "withdraw" functionality
+    // Funds can only be retrived through an action
+    #[payable("*")]
+    #[endpoint(depositTokensForAction)]
+    fn deposit_tokens_for_action(&self) {
```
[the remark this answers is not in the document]
dorin-iancu: Fixed. Now tokens are deposited for specific proposals, with automatic refunds on cancel.

energy-integration/governance-v2/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+
+            gov_actions.push(gov_action);
+        }
+
+        require!(
+            self.total_gas_needed(&gov_actions) < MAX_GAS_LIMIT_PER_BLOCK,
+            "Actions require too much gas to be executed"
+        );
+
+        let proposer = self.blockchain().get_caller();
```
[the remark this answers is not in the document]
dorin-iancu: Done.

energy-integration/governance-v2/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+        self.total_downvotes(proposal_id)
+            .update(|total_downvotes| *total_downvotes += &user_energy);
+
+        self.downvote_cast_event(&downvoter, proposal_id, &user_energy);
+    }
+
+    /// Queue a proposal for execution.
+    /// This can be done only if the proposal has reached the quorum.
+    /// A proposal is considered successful and ready for queing if
+    /// total_votes - total_downvotes >= quorum
```
psorinionut, line 178: This does not longer apply as the formula changed.

energy-integration/governance-v2/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// and all the required payments were deposited
+    #[endpoint]
+    fn queue(&self, proposal_id: ProposalId) {
+        self.require_caller_not_self();
+        require!(
+            self.get_proposal_status(proposal_id) == GovernanceProposalStatus::Succeeded,
+            "Can only queue succeeded proposals"
+        );
+        require!(
```
psorinionut, line 187: I think this condition must be updated.
There should not be a condition for self.pending_payments_for_proposal. When you make a new proposal you set this value and you clear it only in the clearProposal endpoint. Furthermore, I think we should not have this condition because what happens if a proposal requires no payments, as stated in the check from the depositTokensForProposal endpoint?
Also, if there is a need for a second check, besides self.payments_depositor, I think we should have an AND check (&&) instead of OR (||).

energy-integration/governance-v2/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+        let current_block = self.blockchain().get_block_nonce();
+        self.proposal_queue_block(proposal_id).set(&current_block);
+
+        self.proposal_queued_event(proposal_id, current_block);
+    }
+
+    /// Execute a previously queued proposal.
+    /// This will clear the proposal and unlock the governance tokens.
+    /// Said tokens can then be withdrawn and used to vote/downvote other proposals.
```
psorinionut, line 201: Again, I don't think this applies anymore, as we use energy to vote, instead of governance tokens.

psorinionut (approved):

PR #427, opened 2022-07-22
Pair admin actions

claudiu725:


dex/pair/src/config.rs
```diff
@@ -8,29 +8,24 @@ use super::errors::*;
...
-        self.require_permissions();
+        self.require_caller_has_owner_permissions();
         self.state().set(State::PartialActive);
     }
 
     #[endpoint(setFeePercents)]
     fn set_fee_percent(&self, total_fee_percent: u64, special_fee_percent: u64) {
-        self.require_permissions();
+        self.require_caller_has_owner_or_admin_permissions();
         self.set_fee_percents(total_fee_percent, special_fee_percent);
```
dorin-iancu, line 29: Should probably also add some restrictions for the values themselves, maybe add a MAX_FEE_PERCENTAGE constant, so the user doesn't set penalty to 100% or something.

For now, that constant being the equivalent of 5% should be enough.

dex/pair/src/fee.rs
```diff
@@ -38,14 +39,14 @@ pub trait FeeModule:
 
     #[endpoint(whitelist)]
     fn whitelist_endpoint(&self, address: ManagedAddress) {
-        self.require_permissions();
+        self.require_caller_has_owner_or_admin_permissions();
```
dorin-iancu, line 42: I'm not sure if we want the user to whitelist any address they want. Maybe owner-only here.

dex/pair/src/locking_wrapper.rs
```diff
@@ -2,17 +2,14 @@ elrond_wasm::imports!();
 
 #[elrond_wasm::module]
 pub trait LockingWrapperModule:
-    crate::config::ConfigModule + token_send::TokenSendModule + pausable::PausableModule
+    crate::config::ConfigModule
+    + token_send::TokenSendModule
+    + permissions_module::PermissionsModule
+    + pausable::PausableModule
 {
-    #[endpoint(setLockingDeadlineEpoch)]
```
dorin-iancu, line 7: Why was this and `set_unlock_epoch` removed?

claudiu725: According to [the spreadsheet specs](https://docs.google.com/spreadsheets/d/19LhEizrR9nlyPv9yndR83QQZX6FT_03-HhNP1FeLUGg/edit) - both endpoints are not used and should be removed.

dex/pair/src/fee.rs
claudiu725: Done

dex/pair/src/config.rs
claudiu725: Done - set to 5%

dex/pair/src/lib.rs
```diff
@@ -83,9 +87,16 @@ pub trait Pair<ContractReader>:
         self.initial_liquidity_adder()
             .set(&initial_liquidity_adder.into_option());
```
dorin-iancu, line 88: Since it's not possible to skip this argument while also giving admins, I suggest using `ManagedAddress` for argument, and storing `Option::None` if the argument is the zero address. (since this requires the least code changes).

Additionally, make sure you change the code in the `router` contract that deploys pair contracts as well.

dex/farm/src/lib.rs
```diff
@@ -52,7 +53,8 @@ pub trait Farm:
         farming_token_id: TokenIdentifier,
         division_safety_constant: BigUint,
         pair_contract_address: ManagedAddress,
-        mut admins: MultiValueEncoded<ManagedAddress>,
+        owner_opt: OptionalValue<ManagedAddress>,
+        admins: MultiValueEncoded<ManagedAddress>,
```
dorin-iancu, line 57: Same comment as for the Pair contract.

dorin-iancu (approved):

dex/pair/src/lib.rs
claudiu725: Done.

dex/farm/src/lib.rs
claudiu725: Done.

PR #447, opened 2022-09-14
farm with community rewards admin resume feature

psorinionut:


dex/farm_with_community_rewards/src/lib.rs
```diff
@@ -86,8 +88,24 @@ pub trait Farm:
         self.farming_token_id().set(&farming_token_id);
         self.pair_contract_address().set(&pair_contract_address);
 
-        let caller = self.blockchain().get_caller();
```
ovidiuolteanu, line 89: You can leave these two lines as well into the pause whitelist. If the proxy_deployer contract will get a "pause_all" endpoint at any point in time, it would be useful to have it whitelisted as well + to have the ability to pause it via the "only_owner" forwarder endpoint.

ovidiuolteanu (changes requested):

dex/farm_with_community_rewards/src/lib.rs
psorinionut: Done.

dorin-iancu (approved):

PR #461, opened 2022-09-21
fixes + migration test

dorin-iancu:
- migration now generates a single token, with a weighted average of the old token's unlock epochs (weight being the unlock percent at the specific epoch)
- test + fixes


common/common_structs/src/lib.rs
```diff
@@ -144,3 +144,17 @@ pub struct LockedAssetTokenAttributesEx<M: ManagedTypeApi> {
...
+    pub fn average_unlock_epoch(&self) -> Epoch {
+        let mut weight_total = 0;
+        let mut weighted_sum = BigUint::<M>::zero();
+        for milestone in &self.unlock_schedule.unlock_milestones {
+            weighted_sum += milestone.unlock_percent * milestone.unlock_epoch;
+            weight_total += milestone.unlock_percent;
+        }
+
+        let weighted_average = weighted_sum / weight_total;
+        unsafe { weighted_average.to_u64().unwrap_unchecked() }
```
CostinCarabas, line 158: Why is `unsafe` needed here?

locked-asset/simple-lock-energy/src/lock_options.rs
```diff
@@ -70,9 +70,9 @@ pub trait LockOptionsModule {
         );
     }
 
-    fn lock_epoch_to_start_of_month(&self, lock_option: Epoch) -> Epoch {
-        let extra_days = lock_option % EPOCHS_PER_MONTH;
-        lock_option - extra_days
+    fn unlock_epoch_to_start_of_month(&self, unlock_epoch: Epoch) -> Epoch {
+        let extra_days = unlock_epoch % EPOCHS_PER_MONTH;
```
CostinCarabas, line 74: This is not very accurate (`EPOCHS_PER_MONTH`). 
Why is start of month important here? Do we need precision?

common/common_structs/src/lib.rs
dorin-iancu: `unwrap_unchecked` is an unsafe function, as it leads to undefined behaviour in case of `None` (`to_u64` return `Option<u64>`), but a weighted average like this will never go over `u64`, so it's mathematically safe.

locked-asset/simple-lock-energy/src/lock_options.rs
dorin-iancu: This is done so, for example, a token that were to unlock on 5th september, one that would unlock on 10th, and one on 12th, will all unlock on 1st, hence have the same `token_nonce` for the locked token (since they'd have the same attributes).

This approximation is done so users don't fill their wallets with tokens.

CostinCarabas (approved):

dorin-iancu:
Converted to draft until the other PRs are merged, easier to solve eventual merge conflicts on this branch.

claudiu725 (approved):

CostinCarabas (dismissed):

claudiu725 (approved):

PR #462, opened 2022-09-21
import proxy modules + remove FarmType

dorin-iancu:
- simple-lock-energy can now also act as proxy for farms/pairs

### DIFF
--- a/locked-asset/simple-lock/src/proxy_farm.rs
+++ b/locked-asset/simple-lock/src/proxy_farm.rs
@@ -3,17 +3,8 @@ elrond_wasm::derive_imports!();
 
 use crate::{error_messages::*, proxy_lp::LpProxyTokenAttributes};
 
-#[derive(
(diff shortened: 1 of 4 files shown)


claudiu725 (approved):

locked-asset/simple-lock/src/proxy_farm.rs
```diff
@@ -153,8 +138,7 @@ pub trait ProxyFarmModule:
             ));
         }
 
-        let farm_address =
-            self.try_get_farm_address(&lp_proxy_token_attributes.lp_token_id, farm_type);
+        let farm_address = self.try_get_farm_address(&lp_proxy_token_attributes.lp_token_id);
         let enter_farm_result = self.call_farm_enter(
```
CostinCarabas, line 142: If `payments` len is 1 (there are no additional farm payments, only the `proxy_lp_payment`), the `additional_farm_payments` will be empty and the SC will call endpoint `call_farm_enter`. It doesn't seem to affect anything, only gas spent redundantly.

PR #467, opened 2022-09-27
Farm with locked rewards impl

psorinionut:


dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+        amount: BigUint) 
+    {
+        if amount == 0 {
+            return;
+        }
+        let token_id = EgldOrEsdtTokenIdentifier::esdt(token_identifier);
+        self.lock_tokens_and_forward(destination_address, token_id, amount);
+    }
+
+    #[storage_mapper("maxLockOption")]
```
dorin-iancu, line 325: mismatch between mapper name and storage key name

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+        base_farm_reward + boosted_yield_rewards
+    }
+
+    fn require_queried(&self) {
+        let caller = self.blockchain().get_caller();
+        let sc_address = self.blockchain().get_sc_address();
+        require!(caller == sc_address, "May only call this function through VM query");
+    }
+
+    fn send_to_lock_contract_non_zero(
```
dorin-iancu, line 312: This function should return the output payment received from `lock_tokens_and_forward`, since you'll need it to propagate it as endpoint result.

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+            Self::get_default_merged_farm_token_attributes,
+            Self::create_farm_tokens_by_merging,
+        );
+
+        let output_farm_token_payment = base_claim_rewards_result.new_farm_token.payment;
+        let rewards_payment = base_claim_rewards_result.rewards;
+        self.send_payment_non_zero(&caller, &output_farm_token_payment);
+        self.send_to_lock_contract_non_zero(caller, rewards_payment.token_identifier.clone(), rewards_payment.amount.clone());
+
+        (output_farm_token_payment, rewards_payment).into()
```
dorin-iancu, line 133: This should return the locked tokens as output instead of the initial unlocked rewards (i.e. should return the output of `send_to_lock_contract_non_zero`, and not `rewards_payment` directly)

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+                &mut farming_token_payment.amount,
+                &base_exit_farm_result.storage_cache.farming_token_id, 
+                &base_exit_farm_result.storage_cache.reward_token_id,
+            );
+        }
+
+        self.send_payment_non_zero(&caller, &farming_token_payment);
+        self.send_to_lock_contract_non_zero(caller, reward_payment.token_identifier.clone(), reward_payment.amount.clone());
+
+        (farming_token_payment, reward_payment).into()
```
dorin-iancu, line 210: Same as for claim_rewards.

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+            .set_if_empty(DEFAULT_MINUMUM_FARMING_EPOCHS);
+        self.burn_gas_limit().set_if_empty(DEFAULT_BURN_GAS_LIMIT);
+        self.pair_contract_address().set(&pair_contract_address);
+    }
+
+    #[payable("*")]
+    #[endpoint(enterFarm)]
+    fn enter_farm_endpoint(&self) -> EnterFarmResultType<Self::Api> {
+        let output_farm_token_payment = self.enter_farm();
+        let caller = self.blockchain().get_caller();
```
dorin-iancu, line 82: Missing `require_sc_whitelisted` check, I believe.

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+        self.minimum_farming_epochs()
+            .set_if_empty(DEFAULT_MINUMUM_FARMING_EPOCHS);
+        self.burn_gas_limit().set_if_empty(DEFAULT_BURN_GAS_LIMIT);
+        self.pair_contract_address().set(&pair_contract_address);
+    }
+
+    #[payable("*")]
+    #[endpoint(enterFarm)]
+    fn enter_farm_endpoint(&self) -> EnterFarmResultType<Self::Api> {
+        let output_farm_token_payment = self.enter_farm();
```
dorin-iancu, line 81: Shouldn't `enter_farm` also have a `caller` argument?

dex/farm/src/base_functions.rs
```diff
@@ -0,0 +1,213 @@
+#![allow(clippy::too_many_arguments)]
```
dorin-iancu, line 1: format this file

dex/farm-with-locked-rewards/src/lib.rs
psorinionut: Done.

dex/farm/src/base_functions.rs
psorinionut: Done.

dex/farm-with-locked-rewards/src/lib.rs
psorinionut: For the enter farm algorithm we do not need the caller in order to compute the farming position, as the actual send of tokens towards the user does not happen in the base function. 
We will probably need the caller as an argument when the events module will be implemented, but it isn't in this PR.

dorin-iancu (approved):

PR #472, opened 2022-09-30
Token merging

dorin-iancu:
Allow merging of multiple tokens under a single one. This is done through a weighted average based on token amount, and original unlock epoch.

The final unlock epoch is estimated towards the upper bound instead of the usual floor -> start of month. This is done to prevent exploits where merging would be used to keep reducing the lock period indefinitely.

### DIFF
--- a/locked-asset/simple-lock-energy/src/token_merging.rs
+++ b/locked-asset/simple-lock-energy/src/token_merging.rs
@@ -119,6 +119,11 @@ pub trait TokenMergingModule:
                     output_pair.merge_with(amount_attr_pair);
                 }
 
+                let normalized_unlock_epoch = self.unlock_epoch_to_start_of_month_upper_estimate(
+                    output_pair.attributes.unlock_epoch,
+                );
+                output_pair.attributes.unlock_epoch = normalized_unlock_epoch;
+
                 energy.add_after_token_lock(
                     &output_pair.token_amount,
                     output_pair.attributes.unlock_epoch,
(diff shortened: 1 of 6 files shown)


locked-asset/simple-lock-energy/src/token_merging.rs
```diff
@@ -0,0 +1,144 @@
...
+static CANNOT_MERGE_ERR_MSG: &[u8] = b"Cannot merge";
+
+pub trait Mergeable {
+    fn can_merge_with(&self, other: &Self) -> bool;
+
+    fn merge_with(&mut self, other: Self);
+}
+
+impl<M: ManagedTypeApi> Mergeable for EsdtTokenPayment<M> {
+    fn can_merge_with(&self, other: &Self) -> bool {
```
psorinionut, line 16: Why assign variables and not just compare the 2 values?

self.token_identifier == other.token_identifier && self.token_nonce == other.token_nonce

locked-asset/simple-lock-energy/src/token_merging.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+pub struct LockedAmountAttributesPair<M: ManagedTypeApi> {
+    pub token_amount: BigUint<M>,
+    pub attributes: LockedTokenAttributes<M>,
+}
+
+impl<M: ManagedTypeApi> Mergeable for LockedAmountAttributesPair<M> {
+    fn can_merge_with(&self, other: &Self) -> bool {
```
psorinionut, line 38: Same as above

psorinionut (commented):
Small suggestions for code optimization below

locked-asset/simple-lock-energy/src/token_merging.rs
```diff
@@ -0,0 +1,144 @@
...
+                        attributes,
+                    };
+                    output_pair.merge_with(amount_attr_pair);
+                }
+
+                energy.add_after_token_lock(
+                    &output_pair.token_amount,
+                    output_pair.attributes.unlock_epoch,
+                    current_epoch,
+                );
```
claudiu725, line 126: The energy added here uses `output_pair.attributes.unlock_epoch` and not the `normalized_unlock_epoch` which is used below in `self.lock_and_send(&caller, simulated_lock_payment, normalized_unlock_epoch);`
Shouldn't the same `normalized_unlock_epoch` be used in both places?

claudiu725 (approved):

psorinionut (approved):

PR #474, opened 2022-10-04
LKMEX transfer contract

alyn509:


locked-asset/lkmex-transfer/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+    #[payable("*")]
+    #[endpoint(lockFunds)]
+    fn lock_funds(&self, address: ManagedAddress) {
+        let caller = self.blockchain().get_caller();
+        self.check_address_on_cooldown(&caller);
+
+        let payments = self.call_value().all_esdt_transfers();
+        for payment in payments.iter() {
+            require!(
+                payment.token_identifier == TokenIdentifier::from(LPMEX_TOKEN_ID),
```
dorin-iancu, line 52: `LPMEX_TOKEN_ID` should be a token ID in storage, not a constant.

locked-asset/lkmex-transfer/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+                CALLER_ON_COOLDOWN
+            )
+        }
+    }
+
+    #[storage_mapper("locked_funds")]
+    fn locked_funds(
+        &self,
+        owner: &ManagedAddress,
+    ) -> SingleValueMapper<(ManagedVec<EsdtTokenPayment<Self::Api>>, u64)>;
```
dorin-iancu, line 78: Please avoid tuples, use a struct.

locked-asset/lkmex-transfer/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+elrond_wasm::imports!();
+elrond_wasm::derive_imports!();
+
+pub mod constants;
+use crate::constants::*;
+
+#[elrond_wasm::contract]
+pub trait LkmexTransfer {
+    #[init]
+    fn init(&self, unlock_transfer_time: u64, address_cooldown_duration: u64) {
```
dorin-iancu, line 13: Rename `address_cooldown_duration` to something like `epochs_cooldown_duration`, or similar, to make it obvious it's epochs, and not something else (like blocks, timestamp, etc).

Same in other places it's being used.

locked-asset/lkmex-transfer/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    fn check_address_on_cooldown(&self, address: &ManagedAddress) {
+        if !self.address_last_transfer_epoch(address).is_empty() {
+            let current_epoch = self.blockchain().get_block_epoch();
+            let address_cooldown_duration = self.address_cooldown_duration().get();
+            let last_cooldown = self.address_last_transfer_epoch(address).get();
+            require!(
+                current_epoch - last_cooldown > address_cooldown_duration,
+                CALLER_ON_COOLDOWN
+            )
```
dorin-iancu, line 70: This makes the intention clearer:
```suggestion
            let last_transfer_epoch = self.address_last_transfer_epoch(address).get();
            let epochs_since_last_transfer = current_epoch - last_cooldown;
            require!(
                epochs_since_last_transfer > address_cooldown_duration,
                CALLER_ON_COOLDOWN
            )
```

locked-asset/lkmex-transfer/src/lib.rs
dorin-iancu: To help with this, maybe also have an alias like `type Epoch = u64`, and use that instead of u64, to help with code readability.

locked-asset/lkmex-transfer/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+        let caller = self.blockchain().get_caller();
+        self.check_address_on_cooldown(&caller);
+
+        let payments = self.call_value().all_esdt_transfers();
+        for payment in payments.iter() {
+            require!(
+                payment.token_identifier == self.locked_token().get(),
+                BAD_LOCKING_TOKEN
+            )
+        }
```
dorin-iancu, line 72: Rewrite so the locked token ID is only read once from storage:
```suggestion
        let locked_token_id = self.locked_token().get();
        for payment in payments.iter() {
            require!(
                payment.token_identifier == locked_token_id,
                BAD_LOCKING_TOKEN
            )
        }
```

dorin-iancu (changes requested):

dorin-iancu (approved):

PR #476, opened 2022-10-05
proxy lp update

dorin-iancu:


common/modules/utils/src/lib.rs
```diff
@@ -24,13 +26,35 @@ pub trait UtilsModule {
...
+        require!(!payments.is_empty(), ERR_EMPTY_PAYMENTS);
 
         payments
     }
 
+    fn pop_first_payment(
+        &self,
+        payments: &mut PaymentsVec<Self::Api>,
+    ) -> EsdtTokenPayment<Self::Api> {
+        require!(!payments.is_empty(), ERR_EMPTY_PAYMENTS);
```
CostinCarabas, line 50: Why not use `get_non_empty_payments` here?

common/traits/fixed-supply-token/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub trait FixedSupplyToken<M: ManagedTypeApi> {
+    fn get_total_supply(&self) -> &BigUint<M>;
+
+    fn into_part(self, payment_amount: &BigUint<M>) -> Self;
+
+    /// full_value * current_supply / total_supply
+    fn rule_of_three(&self, current_supply: &BigUint<M>, full_value: &BigUint<M>) -> BigUint<M> {
+        let total_supply = self.get_total_supply();
+        if current_supply == total_supply {
```
CostinCarabas, line 13: Why the extra verification? Line 14 would result in the same value.

common/traits/fixed-supply-token/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    /// full_value * current_supply / total_supply
+    fn rule_of_three_non_zero_result(
+        &self,
+        current_supply: &BigUint<M>,
+        full_value: &BigUint<M>,
+    ) -> BigUint<M> {
+        let result = self.rule_of_three(current_supply, full_value);
+        if result == 0 {
```
CostinCarabas, line 27: Shouldn't this behaviour be the default one?

common/traits/mergeable/src/lib.rs
```diff
@@ -6,14 +6,27 @@ pub static CANNOT_MERGE_ERR_MSG: &[u8] = b"Cannot merge";
 
 pub trait Mergeable<M: ManagedTypeApi> {
     fn error_if_not_mergeable(&self, other: &Self) {
-        if !self.can_merge_with(other) {
-            M::error_api_impl().signal_error(CANNOT_MERGE_ERR_MSG);
+        if !self.can_merge_with(&other) {
```
CostinCarabas, line 9: Clippy

common/traits/mergeable/src/lib.rs
```diff
@@ -6,14 +6,27 @@ pub static CANNOT_MERGE_ERR_MSG: &[u8] = b"Cannot merge";
```
CostinCarabas, line 6: Maybe an ERR_MSG file would be useful?

locked-asset/proxy_dex/src/proxy_common.rs
```diff
@@ -8,20 +8,81 @@ use crate::{
...
-pub trait ProxyCommonModule: token_send::TokenSendModule {
+pub trait ProxyCommonModule {
+    fn require_exactly_one_locked<'a>(
+        &self,
+        first_payment: &'a EsdtTokenPayment,
+        second_payment: &'a EsdtTokenPayment,
+    ) -> LockedUnlockedTokenRefPair<'a, Self::Api> {
+        let token_mapper = self.locked_token_ids();
+        let first_is_locked = token_mapper.contains(&first_payment.token_identifier);
+        let second_is_locked = token_mapper.contains(&first_payment.token_identifier);
```
CostinCarabas, line 33: first_payment - > second_payment?

locked-asset/proxy_dex/src/proxy_common.rs
```diff
@@ -8,20 +8,81 @@ use crate::{
...
+        &self,
+        first_payment: &'a EsdtTokenPayment,
+        second_payment: &'a EsdtTokenPayment,
+    ) -> BaseAssetOtherTokenRefPair<'a, Self::Api> {
+        let base_asset_token_id = self.asset_token_id().get();
+        let is_first_token = first_payment.token_identifier == base_asset_token_id;
+        let is_second_token = second_payment.token_identifier == base_asset_token_id;
+
+        if is_first_token {
+            require!(!is_second_token, INVALID_PAYMENTS_ERR_MSG);
```
CostinCarabas, line 62: Is there a more specific error message than `INVALID_PAYMENTS_ERR_MSG`?

locked-asset/proxy_dex/src/sc_whitelist.rs
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    #[only_owner]
+    #[endpoint(removeIntermediatedPair)]
+    fn remove_intermediated_pair(&self, pair_address: ManagedAddress) {
+        self.require_is_intermediated_pair(&pair_address);
+        let _ = self.intermediated_pairs().swap_remove(&pair_address);
+    }
+
+    fn require_is_intermediated_pair(&self, address: &ManagedAddress) {
```
CostinCarabas, line 18: `require_is_intermediated_pair` -> `is_intermediated_pair` ?

locked-asset/proxy_dex/src/wrapped_lp_attributes.rs
```diff
@@ -1,12 +1,117 @@
...
+    pub payment: EsdtTokenPayment<M>,
+    pub attributes: WrappedLpTokenAttributes<M>,
+}
+
+impl<M: ManagedTypeApi + StorageMapperApi + CallTypeApi> WrappedLpToken<M> {
+    pub fn new_from_payments(
+        payments: &PaymentsVec<M>,
+        wrapped_token_mapper: &NonFungibleTokenMapper<M>,
+    ) -> ManagedVec<M, Self> {
+        wrapped_token_mapper.require_all_same_token(&payments);
```
CostinCarabas, line 70: clippy

locked-asset/proxy_dex/src/wrapped_lp_attributes.rs
```diff
@@ -1,12 +1,117 @@
...
+/// The caller function should handle the minting of the new wrapped LP tokens.
+pub fn merge_wrapped_lp_tokens_through_factory<M: CallTypeApi>(
+    factory_address: ManagedAddress<M>,
+    mut wrapped_lp_tokens: ManagedVec<M, WrappedLpToken<M>>,
+) -> WrappedLpTokenAttributes<M> {
+    let first_item = wrapped_lp_tokens.get(0);
+    wrapped_lp_tokens.remove(0);
+
+    let mut merged_wrapped_lp_attributes =
+        first_item.attributes.into_part(&first_item.payment.amount);
```
CostinCarabas, line 98: Should there be a check for lp token_identifier? In `into_part` function you don't check for that.

claudiu725 (changes requested):

common/modules/utils/src/lib.rs
dorin-iancu: Payments is an argument here, so it modifies a given payment Vec.

common/traits/fixed-supply-token/src/lib.rs
dorin-iancu: Since this is the most common case (rarely do people actually split positions), it acts as a minor optimization (comparison instead of multiplication and division on potentially very big numbers)

common/traits/fixed-supply-token/src/lib.rs
dorin-iancu: In most cases, yeah, you don't want a result of 0, but since this is a trait, I've kept the option there.

common/traits/mergeable/src/lib.rs
dorin-iancu: Fixed.

common/traits/mergeable/src/lib.rs
dorin-iancu: Eh, if it was a smart contract, maybe, but since this is a standalone module, I'd rather each keep their own error message, then a big file with all the errors combined.

locked-asset/proxy_dex/src/sc_whitelist.rs
dorin-iancu: `is_intermediated_pair` would generally be a function that returns boolean, while the `require` prefix signals that an error will be thrown if the condition is not fulfilled.

locked-asset/proxy_dex/src/wrapped_lp_attributes.rs
dorin-iancu: Fixed.

locked-asset/proxy_dex/src/proxy_common.rs
dorin-iancu: The message could be something like "Expected at least one token as base asset", but keeping this somewhat generic message helps with code size, since instead of having many different static strings, we have a single one, used from multiple places.

CostinCarabas (approved):

claudiu725 (approved):

sasurobert (approved):

PR #492, opened 2022-10-12
Energy factory week fix

CostinCarabas:


sasurobert (approved):

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -179,7 +179,14 @@ pub trait WeeklyRewardsSplittingModule:
             Energy::default()
         };
 
-        let prev_week = current_week - 1;
+        // Update Current/Previous Week
+        if current_week != self.current_active_week().get() {
```
psorinionut, line 183: Maybe use the mapper like this:
`let current_global_active_week_mapper = self.current_active_week();`
Then you can only use
`current_active_week_mapper.get()` and `current_active_week_mapper.set()` without instantiating the mapper every time.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -179,7 +179,14 @@ pub trait WeeklyRewardsSplittingModule:
...
+            let previous_current_week = self.current_active_week().get();
+            self.last_active_week().set(previous_current_week);
+            self.current_active_week().set(current_week)
+        }
+
+        let prev_week = self.last_active_week().get();
```
psorinionut, line 189: Same as above, as you use `self.last_active_week` more than once.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
CostinCarabas: Nice catch, thanks!

psorinionut (approved):

PR #495, opened 2022-10-14
boosted yields computation based on token nonce

psorinionut:
- Boosted yields rewards are now computed for each farm token nonce. This is so that users with multiple farm positions to properly receive their corresponding rewards.


dex/farm/src/base_functions.rs
```diff
@@ -99,25 +104,56 @@ pub trait BaseFunctionsModule:
...
+        let token_mapper = self.farm_token();
+        let mut output_attributes: FarmTokenAttributes<Self::Api> =
+            self.get_attributes_as_part_of_fixed_supply(&first_payment, &token_mapper);
+        token_mapper.nft_burn(first_payment.token_nonce, &first_payment.amount);
+
+        for payment in &payments {
+            let attributes: FarmTokenAttributes<Self::Api> =
+                self.get_attributes_as_part_of_fixed_supply(&payment, &token_mapper);
+            output_attributes.merge_with(attributes);
+        }
```
dorin-iancu: Seeing as this code is pretty common, rewrite it as method in the utils module, that receives a `PaymentsVec` as argument, and returns `T`, where `T` is the attributes type. Similar to the current `merge_from_payments` implementation, but without any `base_attributes` argument.

I'd say rename the current `merge_from_payments` to `merge_attributes_with_payments`, and name the new one `merge_from_payments`.

dex/farm/src/base_functions.rs
```diff
@@ -99,25 +104,56 @@ pub trait BaseFunctionsModule:
                 &base_exit_farm_result.storage_cache.reward_token_id,
             );
         }
-
         (farming_token_payment, reward_payment).into()
     }
 
+    fn merge_farm_tokens_with_energy(&self, caller: &ManagedAddress) -> EsdtTokenPayment<Self::Api> {
```
dorin-iancu: This should not be a separate method at all. It's basically a copy of the normal merging function, with a call to update energy added at the end.

dex/farm/src/lib.rs
```diff
@@ -155,16 +156,27 @@ pub trait Farm:
...
+    fn merge_farm_tokens_with_energy_endpoint(&self) -> EsdtTokenPayment<Self::Api> {
+        let caller = self.blockchain().get_caller();
+        let new_tokens = self.merge_farm_tokens_with_energy(&caller);
+        self.send_payment_non_zero(&caller, &new_tokens);
+        new_tokens
+    }
+
     #[payable("*")]
     #[endpoint(mergeFarmTokens)]
     fn merge_farm_tokens_endpoint(&self) -> EsdtTokenPayment<Self::Api> {
```
dorin-iancu: There should not be two merge endpoints at all. The old endpoint is enough, with the OptionalArg of original caller. You would need to update energy in both cases either way.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -6,7 +6,7 @@ elrond_wasm::derive_imports!();
 pub mod events;
```
dorin-iancu, line 6: This module should NOT care about farm token nonces, amounts, anything. This is made to be as generic as possible. It's just a module that splits rewards, per week. We should not pollute this module just because we need something very specific for the farm boosted yields.

Instead, we should have a trait (similar to the one from farm), with more or less the following methods:
- fn calculate_rewards_for_week(&sc, &user, week, &energy);
- fn collect_rewards_for_week(&sc, week);
- fn get_claim_progress(&sc, &user) -> ClaimProgress;
- fn set_claim_progress(&sc, &user, &ClaimProgress);

The handling of farm tokens and such should fall into the wrapper that would implement this trait. Easiest would probably be to have a simple module that holds input payments as a static variable, and provides methods for initializing and accessing those. Then you'd have access to them in a fairly efficient way inside your trait's implementation.

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -150,18 +150,19 @@ pub trait Farm:
...
             &attributes,
             &storage_cache,
         )
     }
 
     #[payable("*")]
-    #[endpoint(mergeFarmTokens)]
-    fn merge_farm_tokens_endpoint(&self) -> EsdtTokenPayment<Self::Api> {
+    #[endpoint(mergeFarmTokensWithEnergy)]
+    fn merge_farm_tokens_with_energy_endpoint(&self) -> EsdtTokenPayment<Self::Api> {
```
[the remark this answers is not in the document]
psorinionut: Deleted.

dex/farm/src/base_functions.rs
psorinionut: Done.

dex/farm/src/base_functions.rs
psorinionut: Deleted.

dex/farm/src/lib.rs
psorinionut: Done.

common/modules/utils/src/lib.rs
```diff
@@ -68,6 +68,23 @@ pub trait UtilsModule {
...
+        mapper: &NonFungibleTokenMapper<Self::Api>,
+    ) -> T {
+        let first_payment = self.pop_first_payment(payments);
+        let base_attributes: T =
+            self.get_attributes_as_part_of_fixed_supply(&first_payment, &mapper);
+            mapper.nft_burn(first_payment.token_nonce, &first_payment.amount);
+
+        let output_attributes = self.merge_attributes_from_payments(base_attributes, &payments, &mapper);
+
+        self.burn_multi_esdt(&payments);
```
dorin-iancu, line 82: Should not burn inside this method, as the caller might still need the payments. Let the caller handle the token burning.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -143,20 +156,23 @@ pub trait WeeklyRewardsSplittingModule:
         }
     }
 
+    // !!! TODO  - update user boosted rewards formula
     fn get_user_rewards_for_week(
```
sasurobert, line 160: I agreed with what @dorin-iancu wrote. As only this function need to be rewritten in a separate special trait.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -272,13 +288,110 @@ pub trait WeeklyRewardsSplittingModule:
         )
     }
 
+    // Clears the claim progress for all the additional payment tokens
+    // in order to be able to merge the users claim progress under a single token nonce
+    fn clear_payments_claim_progress(
```
sasurobert, line 293: is this really needed ? and here ? like in rewards-split  - not a good place.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
sasurobert: I pretty much agree with this.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -272,13 +288,110 @@ pub trait WeeklyRewardsSplittingModule:
...
+                require!(
+                    first_payment_claim_week == payment_claim_week,
+                    "The claim week of the sent tokens do not match"
+                );       
+            }
+        }
+        return all_payments.get(0).token_nonce;
+    }
+
+    fn update_user_claim_progress(
```
sasurobert, line 317: this should be in the farm as well.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -272,13 +288,110 @@ pub trait WeeklyRewardsSplittingModule:
...
+            }
+            OptionalValue::None => {
+                let new_claim_progress = self.new_claim_progress_for_user(user);
+                self.current_claim_progress(user, new_nonce)
+                    .set(new_claim_progress);
+            }
+        }
+    }
+
+    fn get_claim_progress_or_default(
```
sasurobert, line 344: this as well - put in farm.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -272,13 +288,110 @@ pub trait WeeklyRewardsSplittingModule:
...
+        if current_claim_mapper.is_empty() {
+            claim_progress = self.new_claim_progress_for_user(user);
+        } else {
+            claim_progress = current_claim_mapper.get();
+        }
+
+        claim_progress
+    }
+
+    fn get_claim_progress_and_clear(
```
sasurobert, line 360: put in farm

sasurobert (commented):
clippy issues.

dex/farm/src/lib.rs
```diff
@@ -155,16 +156,27 @@ pub trait Farm:
...
             &attributes,
             &storage_cache,
         )
     }
 
+    #[payable("*")]
+    #[endpoint(mergeFarmTokensWithEnergy)]
+    fn merge_farm_tokens_with_energy_endpoint(&self) -> EsdtTokenPayment<Self::Api> {
+        let caller = self.blockchain().get_caller();
+        let new_tokens = self.merge_farm_tokens_with_energy(&caller);
```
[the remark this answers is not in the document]
psorinionut: Deleted.

dex/farm/src/lib.rs
```diff
@@ -155,16 +156,27 @@ pub trait Farm:
...
+        self.send_payment_non_zero(&caller, &new_tokens);
+        new_tokens
+    }
+
     #[payable("*")]
     #[endpoint(mergeFarmTokens)]
     fn merge_farm_tokens_endpoint(&self) -> EsdtTokenPayment<Self::Api> {
         let caller = self.blockchain().get_caller();
+        self.require_sc_address_whitelisted(&caller);
```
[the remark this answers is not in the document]
psorinionut: Updated.

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -146,18 +145,18 @@ pub trait Farm:
...
+            farm_token_nonce,
             &farm_token_amount,
             &attributes,
             &storage_cache,
         )
     }
 
     #[payable("*")]
-    #[endpoint(mergeFarmTokens)]
+    #[endpoint(mergeFarmTokensWithEnergy)]
```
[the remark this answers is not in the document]
psorinionut: Updated, we now have only one endpoint.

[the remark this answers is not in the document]
sasurobert: Rename it to mergeFarmTokens - as it was before

common/modules/utils/src/lib.rs
```diff
@@ -67,7 +67,28 @@ pub trait UtilsModule {
         attr.into_part(&payment.amount)
     }
 
-    fn merge_from_payments<T: FixedSupplyToken<Self::Api> + Mergeable<Self::Api> + TopDecode>(
+    fn merge_from_payments_and_burn<
```
sasurobert, line 70: it this used ?

psorinionut: Yes, it is used in the merge_farm_tokens function. As it is a more common code, it was moved as a generic function in the utils modules, to make the contract more readable.

common/modules/farm/farm_base_impl/src/base_traits_impl.rs
```diff
@@ -59,6 +59,7 @@ pub trait FarmContract {
     fn calculate_rewards(
         _sc: &Self::FarmSc,
         _caller: ManagedAddress<<Self::FarmSc as ContractBase>::Api>,
+        _farm_token_nonce: Nonce,
```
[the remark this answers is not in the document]
psorinionut: It is marked as unneeded through "_". As this function is also used for the boosted yields, it needs to have the same number of arguments, even if it is overwritten or not, in order to compile.

dorin-iancu (approved):

sasurobert (approved):

PR #517, opened 2022-10-21
collect undistributed boosted rewards feature

psorinionut:


energy-integration/farm-boosted-yields/src/lib.rs
```diff
@@ -83,6 +86,30 @@ pub trait FarmBoostedYieldsModule:
         self.boosted_yields_factors().set(factors);
     }
 
+    #[endpoint(collectUndistributedBoostedRewards)]
```
sasurobert, line 89: please add tests before merging.

energy-integration/farm-boosted-yields/src/lib.rs
```diff
@@ -3,12 +3,15 @@
...
 
+use core::cmp;
+
 use common_types::{Nonce, PaymentsVec};
 use week_timekeeping::Week;
 use weekly_rewards_splitting::{base_impl::WeeklyRewardsSplittingTraitsModule, ClaimProgress};
 
 const MAX_PERCENT: u64 = 10_000;
 const BLOCKS_PER_WEEK: u64 = 100_800u64;
+pub const COLLECT_UNDISTRIBUTED_BOOSTED_REWARDS_WEEK_OFFSET: usize = 4;
```
sasurobert, line 14: let's be safe and change to 5.

sasurobert (changes requested):

energy-integration/farm-boosted-yields/src/lib.rs
psorinionut: Added.

energy-integration/farm-boosted-yields/src/lib.rs
psorinionut: The algorithm uses this constant in an exclusive manner, meaning that the rewards cannot be collected if the number of passed weeks is not exclusively greater than 4 (as shown in the test). Changing it to 5 would result in a claim logic inconsistency.

sasurobert (approved):

PR #544, opened 2022-11-11
token unlock with unbond period

psorinionut:
- Tokens now have an unbonding period upon unlocking
- Added an extra test for the farm boosted rewards computation

### DIFF
--- a/common/modules/token_unstake/src/token_unstake.rs
+++ b/common/modules/token_unstake/src/token_unstake.rs
@@ -3,6 +3,8 @@
 elrond_wasm::imports!();
 elrond_wasm::derive_imports!();
 
+pub const DEFAULT_UNBOND_EPOCHS: u64 = 10;
+
 #[derive(ManagedVecItem, TopEncode, TopDecode, NestedEncode, NestedDecode, TypeAbi, Clone)]
 pub struct UnstakePair<M: ManagedTypeApi> {
     pub unlock_epoch: u64,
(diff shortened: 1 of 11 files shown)


locked-asset/energy-factory/src/lib.rs
```diff
@@ -161,13 +165,23 @@ pub trait SimpleLockEnergy:
...
-            &output_payment.token_identifier,
+
+        let token_payment = EsdtTokenPayment::new(
+            output_payment.token_identifier.clone(),
             0,
-            &output_payment.amount,
+            output_payment.amount.clone(),
         );
 
+        let unbond_epochs = self.unbond_epochs().get();
```
sasurobert, line 175: we do not need to lock for unbonding if FEEs are 0.

common/modules/token_unstake/src/token_unstake.rs
```diff
@@ -0,0 +1,55 @@
...
+pub const DEFAULT_UNBOND_EPOCHS: u64 = 10;
+
+#[derive(ManagedVecItem, TopEncode, TopDecode, NestedEncode, NestedDecode, TypeAbi, Clone)]
+pub struct UnstakePair<M: ManagedTypeApi> {
+    pub unlock_epoch: u64,
+    pub token_payment: EsdtTokenPayment<M>,
+}
+
+#[elrond_wasm::module]
+pub trait TokenUnstakeModule: token_send::TokenSendModule {
```
sasurobert, line 15: move to a new contract.

locked-asset/energy-factory/src/lib.rs
psorinionut: Updated.

common/modules/token_unstake/src/token_unstake.rs
psorinionut: Will be done in another PR.

PR #548, opened 2022-11-12
token unstake contract

psorinionut:
- Tokens now have an unbond period after `unlock` or `unlockEarly`
- unbonding is done through a separate contract. The unbond process can also be cancelled, in which case, the user receives their locked tokens back
- tests for unbond cancel to come in a future PR, this one is big enough already
- also, another PR will move the whole fees collector logic into the `unstake` SC, as it's more tied up to that now


locked-asset/token-unstake/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+#[derive(ManagedVecItem, TopEncode, TopDecode, NestedEncode, NestedDecode, TypeAbi, Clone)]
+pub struct UnstakePair<M: ManagedTypeApi> {
+    pub unlock_epoch: u64,
+    pub token_payment: EsdtTokenPayment<M>,
+}
+
+#[elrond_wasm::contract]
+pub trait TokenUnstakeModule: token_send::TokenSendModule {
+    #[init]
+    fn init(&self) {
```
dorin-iancu, line 17: Since this is now a separate contract (and we don't pollute the already heavy `init` of `energy-factory`, I'd say we can add `unbond_epochs` as init argument, and remove the default const.

locked-asset/token-unstake/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+        let current_epoch = self.blockchain().get_block_epoch();
+        let unbond_epochs = self.unbond_epochs().get();
+        let unstake_pair = UnstakePair {
+            unlock_epoch: current_epoch + unbond_epochs,
+            token_payment,
+        };
+        self.unlocked_tokens_for_user(&caller)
+            .update(|unstake_pairs| {
+                unstake_pairs.push(unstake_pair);
+            });
```
dorin-iancu, line 48: I think we can optimize here. To store the least amount of entries, we should check if the last entry has the same `unlock_epoch`, and if yes, we can simply do `amount += ` on that entry.

locked-asset/token-unstake/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn deposit_user_tokens(&self, caller: &ManagedAddress) {
+        let token_payment = self.call_value().single_esdt();
+        require!(
+            token_payment.token_nonce == 0,
+            "Can only unstake fungible tokens"
+        );
+        require!(
+            token_payment.amount > 0,
+            "Payment amount must be greater than 0"
+        );
```
dorin-iancu, line 38: Not required, since transferring ESDT with 0 amount is no longer possible.

locked-asset/token-unstake/src/lib.rs
dorin-iancu: Additionally, as we discussed, let's add a `WhitelistMapper` for whitelisted tokens, and give a `MultiValueEncoded` of `TokenIdentifier` as argument here. Most likely this will be for one token only initially, but maybe we could reuse this for multiple different contracts.

locked-asset/token-unstake/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+                payments.push(unstake_payment.token_payment);
+            } else {
+                remaining_tokens_for_user.push(unstake_payment);
+            }
+        }
+        if remaining_tokens_for_user.is_empty() {
+            unlocked_tokens_for_user_mapper.clear();
+        } else {
+            unlocked_tokens_for_user_mapper.set(remaining_tokens_for_user);
+        };
```
dorin-iancu, line 70: Can simply `set` here, as if you're setting with an empty vec, it's the same as clearing.

locked-asset/token-unstake/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+
+elrond_wasm::imports!();
+elrond_wasm::derive_imports!();
+
+pub const DEFAULT_UNBOND_EPOCHS: u64 = 10;
+
+#[derive(ManagedVecItem, TopEncode, TopDecode, NestedEncode, NestedDecode, TypeAbi, Clone)]
+pub struct UnstakePair<M: ManagedTypeApi> {
+    pub unlock_epoch: u64,
+    pub token_payment: EsdtTokenPayment<M>,
```
dorin-iancu, line 11: I think we should just use `TokenAmountPair` struct here, as we don't need to keep storing nonces of 0 for every entry.

locked-asset/token-unstake/src/lib.rs
psorinionut: Done.

locked-asset/token-unstake/src/lib.rs
psorinionut: Done.

locked-asset/token-unstake/src/lib.rs
psorinionut: I've kept it as a Payment, to more easily pass around the data, without the need to decompose and then create new payments again when needed.

locked-asset/token-unstake/src/lib.rs
psorinionut: I think the check is a bit more complex as we have a managedVec of unknown size. We need to make a lot of new checks (if the vector is empty, how to get the last item and only update that one and so on), and I don't know if it is worth the extra code for that optimization.

sasurobert (commented):
Only the clippy for now.

locked-asset/energy-factory/tests/energy_factory_setup/mod.rs
```diff
@@ -243,14 +242,14 @@ where
...
 
-    pub fn claim_unlocked_tokens(&mut self, caller: &Address) {
-        let rust_zero = rust_biguint!(0u64);
-        self.b_mock
-            .execute_tx(caller, &self.token_unstake_wrapper, &rust_zero, |sc| {
-                sc.claim_unlocked_tokens();
-            })
-            .assert_ok();
-    }
+    // pub fn claim_unlocked_tokens(&mut self, caller: &Address) {
```
sasurobert, line 245: why commented ?

locked-asset/energy-factory/tests/energy_factory_setup/unbond_sc_mock.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.send().esdt_local_burn(
+            &locked_tokens.token_identifier,
+            locked_tokens.token_nonce,
+            &locked_tokens_burn_amount,
+        );
+
+        let (dest_user, ()) =
+            load_endpoint_args::<DebugApi, (ManagedAddress<DebugApi>, ())>(("dest_user", ()));
+
+        // let dest_user_addr_raw = <Self as ContractBase>::Api::argument_api_impl()
```
sasurobert, line 52: delete commented code.

locked-asset/energy-factory/src/token_merging.rs
```diff
@@ -102,16 +102,29 @@ pub trait TokenMergingModule:
         self.require_not_paused();
 
         let payments = self.get_non_empty_payments();
-        let output_amount_attributes = self.merge_tokens(payments, opt_original_caller);
+        let caller = self.blockchain().get_caller();
+        let original_caller = match opt_original_caller {
```
sasurobert, line 106: we have a common function for this, no ?

locked-asset/energy-factory/src/unlock_with_penalty.rs
```diff
@@ -91,8 +91,23 @@ pub trait UnlockWithPenaltyModule:
         let new_unlock_epoch = current_epoch + reduce_result.new_lock_epochs;
 
         let unlocked_tokens = reduce_result.unlocked_tokens;
+        let penalty_amount = &payment.amount - &unlocked_tokens.amount;
         let new_locked_tokens = self.lock_tokens(unlocked_tokens, new_unlock_epoch);
 
+        let amount_to_burn = &payment.amount - &penalty_amount;
```
sasurobert, line 97: so all of this was missing and was an error here ? does burn_penalty resolve the sending of 50% out of that to the fees_collector ?

locked-asset/energy-factory/tests/energy_factory_setup/unbond_sc_mock.rs
dorin-iancu: Done,

locked-asset/energy-factory/src/token_merging.rs
dorin-iancu: Fixed.

locked-asset/energy-factory/src/unlock_with_penalty.rs
dorin-iancu: No, previously, we were doing this in the common function, but we can't do that anymore. For `unlockEarly`, we don't have to burn the locked token, in case user wants to cancel unbond. For `reduceLockPeriod`, we do have to burn the old token, as we mint a new one with the new lock period.

sasurobert (approved):

PR #563, opened 2022-11-17
Gov: Create WaitForFee status

CostinCarabas:
Proposer doesn't vote by default anymore
Add fee for propose and min_free to propose
Other users can contribute to known proposal
If threshold is reached proposal is Active
If in WaitForFee status, others can claim their deposited tokens.


energy-integration/governance-v2/src/configurable.rs
```diff
@@ -152,4 +171,8 @@ pub trait ConfigurablePropertiesModule: energy_query::EnergyQueryModule {
     #[view(getLockTimeAfterVotingEndsInBlocks)]
     #[storage_mapper("lockTimeAfterVotingEndsInBlocks")]
     fn lock_time_after_voting_ends_in_blocks(&self) -> SingleValueMapper<u64>;
+
+    #[view(getGovernanceTokenId)]
+    #[storage_mapper("governanceTokenId")]
+    fn governance_token_id(&self) -> SingleValueMapper<TokenIdentifier>;
```
dorin-iancu, line 177: Should be `fee_token_id` instead. `governance_token_id` makes me think it's used for voting.

energy-integration/governance-v2/src/lib.rs
```diff
@@ -25,34 +27,76 @@ pub trait GovernanceV2:
...
         );
 
-        let actual_payments = self.call_value().all_esdt_transfers();
+        let additional_fee = self.call_value().single_esdt();
         require!(
-            actual_payments == required_payments,
-            "Invalid payments, must match the required payments"
+            self.governance_token_id()
+                .get()
+                .eq(&additional_fee.token_identifier),
```
dorin-iancu, line 47: Use `==` operator instead of manually calling `eq`

energy-integration/governance-v2/src/lib.rs
```diff
@@ -25,34 +27,76 @@ pub trait GovernanceV2:
...
+            WRONG_TOKEN_ID
         );
 
         let caller = self.blockchain().get_caller();
-        depositor_mapper.set(&caller);
+        let mut proposal = self.proposals().get(proposal_id);
+        proposal.fees.entries.push(FeeEntry {
+            depositor_addr: caller.clone(),
+            tokens: additional_fee.clone(),
```
dorin-iancu, line 55: Should probably also have a minimum amount per deposit, so the SC can't get stuck if someone keeps spamming payments of 0.000000....1 tokens.

energy-integration/governance-v2/src/lib.rs
```diff
@@ -84,24 +128,28 @@ pub trait GovernanceV2:
...
         require!(
             user_energy >= min_energy_for_propose,
             "Not enough energy for propose"
         );
 
+        let user_fee = self.call_value().single_esdt();
+        require!(
+            self.governance_token_id()
+                .get()
+                .eq(&user_fee.token_identifier),
```
dorin-iancu, line 141: ==

energy-integration/governance-v2/src/lib.rs
```diff
@@ -25,34 +27,76 @@ pub trait GovernanceV2:
...
+                    .fees
+                    .entries
+                    .iter()
+                    .position(|fee_entry| fee_entry.depositor_addr == caller)
+                    .unwrap();
+                proposal.fees.entries.remove(index);
+
+                self.proposals().set(proposal_id, &proposal);
+                self.user_claim_deposited_tokens_event(&caller, proposal_id, &payment);
+                return;
```
dorin-iancu, line 97: There are a couple problems with this:
- first, this is a very inefficient way of doing this. First, we go over the vec one by one, then we perform an additional search for the item again. 
- second, we assume each user has only one position, which is not enforced

Rewrite this using a `while` loop, something like:
let mut i = 0;
while i < vec.len() {
     if saved_user == caller { 
          vec.remove(i);
          // save payment to another vec, to send at the end 
      }
     else { i+= 1; } 
}

energy-integration/governance-v2/src/lib.rs
```diff
@@ -112,26 +160,26 @@ pub trait GovernanceV2:
             "Actions require too much gas to be executed"
         );
 
+        let mut fees_entries = ManagedVec::new();
+        fees_entries.push(FeeEntry {
+            depositor_addr: proposer.clone(),
+            tokens: user_fee.clone(),
+        });
```
dorin-iancu, line 167: Use `ManagedVec::from_single_item()` instead of `new`, then `push`

energy-integration/governance-v2/src/lib.rs
```diff
@@ -285,14 +323,16 @@ pub trait GovernanceV2:
                 );
             }
             GovernanceProposalStatus::Defeated => {}
+            GovernanceProposalStatus::WaitingForFees => {
+                // TODO: Dacă propunerea e defeated, dăm fee-ul înapoi?
```
dorin-iancu, line 327: Nope, they should all be burned.

energy-integration/governance-v2/src/lib.rs
```diff
@@ -319,26 +359,25 @@ pub trait GovernanceV2:
     }
 
     fn refund_payments(&self, proposal_id: ProposalId) {
-        let payments = self.required_payments_for_proposal(proposal_id).get();
-        if payments.is_empty() {
+        let payments = self.proposals().get(proposal_id).fees;
+        if payments.entries.is_empty() {
             return;
         }
```
dorin-iancu, line 365: No need for this additional check, as the `for` will do this for you.

energy-integration/governance-v2/src/views.rs
```diff
@@ -100,6 +104,10 @@ pub trait ViewsModule:
         proposal_id >= 1 && proposal_id <= self.proposals().len()
     }
 
+    fn proposal_reached_min_fees(&self, proposal_id: ProposalId) -> bool {
+        self.proposals().get(proposal_id).fees.total_amount >= self.min_fee_for_propose().get()
```
dorin-iancu, line 108: It'd a bit cleaner if each comparison member is evaluated separately
```suggestion
        let accumulated_fees = self.proposals().get(proposal_id).fees.total_amount;
        let min_fees = self.min_fee_for_propose().get();
        accumulated_fees >= min_fees
```

energy-integration/governance-v2/src/lib.rs
sasurobert: Please delete comment. Use comment in English always.

sasurobert: Fees do not have to be burnt, but we keep them in the contract. Delete the associated storage for that FEE.

energy-integration/governance-v2/src/lib.rs
sasurobert: Do no resolve conversations which are not yet resolved. You should resolve conversations only after commit is pushed to branch.

sasurobert (changes requested):

energy-integration/governance-v2/src/lib.rs
```diff
@@ -112,26 +159,29 @@ pub trait GovernanceV2:
             "Actions require too much gas to be executed"
         );
 
+        let fees_entries = ManagedVec::from_single_item(FeeEntry {
+            depositor_addr: proposer.clone(),
+            tokens: user_fee.clone(),
+        });
+        // fees_entries.push(FeeEntry {
```
sasurobert, line 166: no commented code please.

sasurobert (approved):

PR #570, opened 2022-11-21
fix farm token leak

dorin-iancu:


psorinionut (approved):

psorinionut (approved):

locked-asset/proxy_dex/src/proxy_farm.rs
```diff
@@ -150,8 +150,8 @@ pub trait ProxyFarmModule:
     #[endpoint(exitFarmProxy)]
     fn exit_farm_proxy(
         &self,
-        exit_amount: BigUint,
         farm_address: ManagedAddress,
+        exit_amount: BigUint,
```
sasurobert, line 154: where is the unit test for this ?

PR #575, opened 2022-11-22
LKMEX Governance

ccorcoveanu:
Changed only needed functionality so that instead of energy:
- user sends the voting power himself (before, energy was queried from the energy SC)
- additionally the user needs to pass a correct merkle proof for that VP
- contract verifies the proof against a stored root hash 
Note: When generating the merkle tree, all hashes need to be sorted so that the verification function is as simple as possible

### DIFF
--- a/energy-integration/governance-v2/meta/Cargo.toml
+++ b/energy-integration/governance-v2/meta/Cargo.toml
@@ -11,4 +11,4 @@ authors = [ "you",]
 path = ".."
 
 [dependencies.elrond-wasm-debug]
-version = "=0.36.1"
+version = "0.36.1"
--- a/energy-integration/governance-v2/meta/Cargo.toml
+++ b/energy-integration/governance-v2/meta/Cargo.toml
@@ -11,4 +11,4 @@ authors = [ "you",]
 path = ".."
 
 [dependencies.elrond-wasm-debug]
-version = "0.36.1"
+version = "=0.36.1"
--- a/energy-integration/governance-v2/src/lib.rs
+++ b/energy-integration/governance-v2/src/lib.rs
@@ -69,6 +69,7 @@ pub trait GovernanceV2:
     ///
     /// Returns the ID of the newly created proposal.
     #[endpoint]
+    #[only_owner]
     fn propose(
         &self,
         root_hash: ManagedByteArray<HASH_LENGTH>,
--- a/energy-integration/governance-v2/src/lib.rs
+++ b/energy-integration/governance-v2/src/lib.rs
@@ -359,14 +359,11 @@ pub trait GovernanceV2:
 
         for proof_item in proof {
             if BigUint::from(hash.as_managed_buffer()) < BigUint::from(proof_item.as_managed_buffer()) {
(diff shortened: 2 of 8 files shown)


energy-integration/governance-v2/meta/Cargo.toml
```diff
@@ -11,4 +11,4 @@ authors = [ "you",]
 path = ".."
 
 [dependencies.elrond-wasm-debug]
-version = "=0.36.1"
+version = "0.36.1"
```
dorin-iancu, line 14: Leave the `=0.36.1` please, this ensures we use that specific version.

energy-integration/governance-v2/src/lib.rs
```diff
@@ -143,7 +139,7 @@ pub trait GovernanceV2:
 
     /// Vote on a proposal. The voting power depends on the user's energy.
     #[endpoint]
-    fn vote(&self, proposal_id: ProposalId, vote: VoteType) {
+    fn vote(&self, proposal_id: ProposalId, vote: VoteType, power: BigUint<Self::Api>, proof: ArrayVec<ManagedByteArray<HASH_LENGTH>, PROOF_LENGTH>) {
```
dorin-iancu, line 142: run `cargo fmt` to format the code

energy-integration/governance-v2/src/lib.rs
```diff
@@ -341,4 +348,29 @@ pub trait GovernanceV2:
 
         self.proposal_votes(proposal_id).clear();
     }
+
+    fn verify_merkle_proof(&self, power: BigUint<Self::Api>, proof: ArrayVec<ManagedByteArray<HASH_LENGTH>, PROOF_LENGTH>, root_hash: ManagedByteArray<HASH_LENGTH>) -> bool {
```
dorin-iancu, line 352: Verification logic should be a separate module

energy-integration/governance-v2/src/lib.rs
```diff
@@ -341,4 +348,29 @@ pub trait GovernanceV2:
...
+                hash = self.crypto().sha256(tst);
+            } else {
+                let mut tst = proof_item.as_managed_buffer().clone();
+                tst.append(hash.as_managed_buffer());
+
+                hash = self.crypto().sha256(tst);
+            }
+        }
+
+        return hash == root_hash;
```
dorin-iancu, line 374:
```suggestion
        hash == root_hash
```

energy-integration/governance-v2/meta/Cargo.toml
ccorcoveanu: done

energy-integration/governance-v2/src/lib.rs
```diff
@@ -71,23 +71,18 @@ pub trait GovernanceV2:
     #[endpoint]
     fn propose(
         &self,
+        root_hash: ManagedByteArray<HASH_LENGTH>,
```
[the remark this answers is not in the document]
ccorcoveanu: added onlu_owner

PR #582, opened 2022-11-23
proxy dex upgrade version

dorin-iancu:


dex/farm/src/lib.rs
```diff
@@ -158,10 +158,12 @@ pub trait Farm:
         let mut context = self.new_farm_context(opt_accept_funds_func);
 
         self.load_state(&mut context);
-        require!(
-            context.get_contract_state().unwrap() == &State::Active,
-            ERROR_NOT_ACTIVE
-        );
+        // Allow exit even if paused
+        //
+        // require!(
```
sasurobert, line 163: we will put into a state - so we need a state.

common/modules/token_send/src/token_send.rs
```diff
@@ -23,6 +23,8 @@ pub trait TokenSendModule {
             function = ManagedBuffer::new();
         }
 
+        self.send().direct_with_gas_limit();
```
sasurobert, line 26: why is this needed ?

dorin-iancu: It isn't, I started updating this module, and eventually gave up and didn't use it in any of the contracts anymore

PR #583, opened 2022-11-23
fees collector update energy after claim rewards

psorinionut:


energy-integration/fees-collector/src/lib.rs
```diff
@@ -37,12 +46,43 @@ pub trait FeesCollector:
...
         let rewards = self.claim_multi(&wrapper, &caller);
+        let energy_token_id = self.locked_token_id().get();
+        let mut require_energy_update = false;
         if !rewards.is_empty() {
             self.send().direct_multi(&caller, &rewards);
+
+            let current_epoch = self.blockchain().get_block_epoch();
+            let mut energy = self.get_energy_entry(&caller);
+            for reward in &rewards {
+                if reward.token_identifier == energy_token_id && reward.token_nonce > 0 {
```
sasurobert, line 57: what if we simplify and give 4 year LKMEX when the user makes a claimReward ? That way we can use the same function the farm contracts are using, that one updates the energy as well. Furthermore, we can simplify the collection of LKMEXs. Today, will make a decision about this.

energy-integration/fees-collector/src/lib.rs
```diff
@@ -37,12 +46,43 @@ pub trait FeesCollector:
...
         }
 
         rewards
     }
+
+    fn set_energy_in_factory(&self, user: ManagedAddress, energy: Energy<Self::Api>) {
+        let sc_address = self.energy_factory_address().get();
+        let _: () = self
+            .energy_factory_proxy(sc_address)
+            .set_user_energy_after_locked_token_transfer(user, energy)
```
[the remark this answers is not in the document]
psorinionut: This was the endpoint that was available (it practically allows a whitelisted contract to set the users energy). I didn't add any more endpoints, to further increase the contract's size.

energy-integration/fees-collector/src/lib.rs
psorinionut: We'll be done in a new PR.

PR #586, opened 2022-11-23
Factory upgrade

dorin-iancu:


locked-asset/factory/src/energy.rs
```diff
@@ -0,0 +1,205 @@
...
+            amount: BigInt::zero(),
+            last_update_epoch: 0,
+            total_locked_tokens: BigUint::zero(),
+        }
+    }
+}
+
+impl<M: ManagedTypeApi> Energy<M> {
+    #[inline]
+    pub fn new(
```
sasurobert, line 25: energy is used ONLY when unlocking LKMEX v1 - I suppose ?

dorin-iancu: Yes, it is only used to update energy after unlock. At lock, energy is not updated.

sasurobert (approved):

PR #589, opened 2022-11-24
fees rewards with predefined lock period

psorinionut:


locked-asset/token-unstake/src/fees_accumulation.rs
```diff
@@ -72,7 +71,8 @@ pub trait FeesAccumulationModule:
         );
 
         let payment = self.call_value().single_esdt();
-        let locked_token_id = self.get_locked_token_id(&energy_factory_addr);
+        let locked_token_id =
```
dorin-iancu, line 74: To better isolate this somewhat low-level code, `get_locked_token_id` should internally use `LOCKED_TOKEN_ID_STORAGE_KEY`, not have it as argument.

locked-asset/token-unstake/src/fees_accumulation.rs
```diff
@@ -120,21 +107,21 @@ pub trait FeesAccumulationModule:
...
-                fees_attributes.attributes.unlock_epoch,
-            )
-            .execute_on_dest_context();
+
+        let fee_tokens: EsdtTokenPayment = self.lock_virtual(
+            self.get_locked_token_id(&energy_factory_addr, BASE_TOKEN_ID_STORAGE_KEY),
+            total_fees,
+            self.blockchain().get_sc_address(),
+            energy_factory_addr,
+        );
```
dorin-iancu, line 124: Fees collector should keep this total amount, and lock_virtual at the start of each claim week, not the unstake contract.

The flow should be:
- user unlocks, X tokens penalty
- unstake SC deposits those tokens in the fees collector SC
- fees collector checks at deposit time, that these are locked_tokens. If yes, it burns them, and adds them to a total, just like the other fungible tokens
- when a user collects rewards, it gives all fungible tokens as normal, but for locked tokens, it then calls virtual_lock

locked-asset/token-unstake/src/fees_accumulation.rs
```diff
@@ -120,21 +107,21 @@ pub trait FeesAccumulationModule:
...
 
         let fees_collector_addr = self.fees_collector_address().get();
         let _: IgnoreValue = self
```
sasurobert, line 127: we do not have to deposit these tokens directly into the fess_collector contract. We might simply tell the fees_collector how much LKMEX it needs to distribute for the users. On claimRewards from fees_collector should the actual NFT create happen.

psorinionut: Updated.

locked-asset/token-unstake/src/fees_accumulation.rs
psorinionut: Done.

locked-asset/token-unstake/src/fees_accumulation.rs
psorinionut: Updated.

energy-integration/fees-collector/src/fees_accumulation.rs
```diff
@@ -25,14 +27,22 @@ pub trait FeesAccumulationModule:
             self.known_tokens().contains(&payment.token_identifier),
```
dorin-iancu, line 27: Should have an `else` clause for payment being locked_token. Alternatively, add locked_token_id in `known_tokens` at init.

energy-integration/fees-collector/src/lib.rs
```diff
@@ -46,33 +47,41 @@ pub trait FeesCollector:
...
+                        energy_factory_addr,
+                    );
+                    let current_epoch = self.blockchain().get_block_epoch();
+                    let mut energy = self.get_energy_entry(&caller);
+                    let attributes: LockedTokenAttributes<Self::Api> = self.get_token_attributes(
+                        &locked_rewards.token_identifier,
+                        locked_rewards.token_nonce,
+                    );
                     if attributes.unlock_epoch > current_epoch {
                         energy.add_after_token_lock(
```
dorin-iancu, line 68: Import this module and use `add_energy_to_destination`
https://github.com/ElrondNetwork/sc-dex-rs/blob/feat/dex_v2/locked-asset/lkmex-transfer/src/energy_transfer.rs

locked-asset/token-unstake/src/fees_accumulation.rs
```diff
@@ -1,12 +1,8 @@
 elrond_wasm::imports!();
```
sasurobert, line 1: should be renamed as it does not accumulate fee's anymore.

sasurobert: when is depositFees called ? Isn't fees taken when user makes the unBond ? are not the tokens already inside this contract ?

energy-integration/fees-collector/src/fees_accumulation.rs
psorinionut: I would not do an else check, as I already have the same check below. I've taken the assumption that the locked_token_id is already whitelisted, but I've also added it in the init function just for safety.

energy-integration/fees-collector/src/lib.rs
```diff
@@ -46,33 +47,41 @@ pub trait FeesCollector:
...
-                if reward.token_identifier == energy_token_id && reward.token_nonce > 0 {
-                    let attributes: LockedTokenAttributes<Self::Api> =
-                        self.get_token_attributes(&reward.token_identifier, reward.token_nonce);
+                if reward.token_identifier == locked_token_id {
+                    let energy_factory_addr = self.energy_factory_address().get();
+                    let locked_rewards = self.lock_virtual(
+                        self.get_base_token_id(&energy_factory_addr),
+                        reward.amount.clone(),
+                        caller.clone(),
+                        energy_factory_addr,
```
sasurobert, line 59: user_address.

energy-integration/fees-collector/src/lib.rs
```diff
@@ -46,33 +47,41 @@ pub trait FeesCollector:
...
+                    );
+                    let current_epoch = self.blockchain().get_block_epoch();
```
sasurobert, line 61: delete next lines

energy-integration/fees-collector/src/lib.rs
```diff
@@ -46,33 +47,41 @@ pub trait FeesCollector:
...
+                            &locked_rewards.amount,
                             attributes.unlock_epoch,
                             current_epoch,
                         );
-
-                        require_energy_update = true;
                     }
+                    self.set_energy_in_factory(caller.clone(), energy);
+                } else {
+                    self.send().direct_esdt(
```
sasurobert, line 76: payments vec - and send with multi transfer ?

energy-integration/fees-collector/src/lib.rs
```diff
@@ -46,33 +47,41 @@ pub trait FeesCollector:
...
+                        &reward.amount,
+                    );
                 }
             }
-            if require_energy_update {
-                self.set_energy_in_factory(caller, energy);
-            }
         }
-
         rewards
```
sasurobert, line 85: return new updated payment vector . and multi transfer.

sasurobert (changes requested):

energy-integration/fees-collector/src/fees_accumulation.rs
psorinionut: Update. It does not matter if you set the locked_token_id in the init function, as add_known_tokens endpoint overwrites the storage anyway. So either way, the locked tokenId must be set when calling this endpoint.

energy-integration/fees-collector/src/lib.rs
psorinionut: No longer needed.

energy-integration/fees-collector/src/lib.rs
psorinionut: Updated.

energy-integration/fees-collector/src/lib.rs
psorinionut: Done.

energy-integration/fees-collector/src/lib.rs
psorinionut: Done.

energy-integration/fees-collector/src/lib.rs
psorinionut: Done.

locked-asset/token-unstake/src/fees_accumulation.rs
psorinionut: 1. Renamed to FeesHandler.
2. depositFees function is called when you reduceLockPeriod. The other case that you may be thinking of, is when the user completely unlocks his tokens (unlock_with_penalty), and in this case, yes, it only applies when the user unbonds the tokens.

dorin-iancu (dismissed):

sasurobert (approved):

PR #594, opened 2022-11-25
new boosted yields formula impl

dorin-iancu:


dex/farm/src/base_functions.rs
```diff
@@ -133,6 +155,10 @@ pub trait BaseFunctionsModule:
         let mut farming_token_payment = base_exit_farm_result.farming_token_payment;
         let reward_payment = base_exit_farm_result.reward_payment;
 
+        self.set_farm_supply_for_current_week(
```
sasurobert, line 158: it this mandatory ? can't you do this down where total_energy_per_week_is_called

sasurobert (approved):

PR #599, opened 2022-11-28
error on save energy for sc

dorin-iancu:


locked-asset/energy-factory/src/migration.rs
```diff
@@ -45,6 +45,10 @@ pub trait SimpleLockMigrationModule:
         original_caller: ManagedAddress,
         epoch_amount_pairs: UnlockEpochAmountPairs<Self::Api>,
     ) {
+        if self.blockchain().is_smart_contract(&original_caller) {
+            return;
```
sasurobert, line 49: you need to return error.

sasurobert (approved):

PR #601, opened 2022-11-28
locked token per block

dorin-iancu:


energy-integration/fees-collector/src/additional_locked_tokens.rs
```diff
@@ -0,0 +1,54 @@
...
+        let current_week = self.get_current_week();
+        if last_update_week == current_week {
+            return;
+        }
+
+        let last_update_block_mapper = self.last_locked_tokens_add_block();
+        let last_block = last_update_block_mapper.get();
+        let current_block = self.blockchain().get_block_nonce();
+
+        let block_diff = current_block - last_block;
```
sasurobert, line 31: no need for block different. have a constant for number of blocks per week. We want to allocate X LKMEX for the Last Week. Exact number, not differences.

psorinionut: Done.

energy-integration/fees-collector/src/additional_locked_tokens.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn set_locked_tokens_per_block(&self, locked_tokens_per_block: BigUint) {
+        self.accumulate_additional_locked_tokens();
+        self.locked_tokens_per_block().set(locked_tokens_per_block);
+    }
+
+    fn accumulate_additional_locked_tokens(&self) {
+        let last_update_week_mapper = self.last_locked_token_add_week();
+        let last_update_week = last_update_week_mapper.get();
+        let current_week = self.get_current_week();
+        if last_update_week == current_week {
```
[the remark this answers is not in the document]
psorinionut: We do not want to terminate the function, as this part only applies if we need to distribute the tokens for the next week. Otherwise, it means that the tokens were already distributed and the function should proceed to allow the user to claim his tokens.

sasurobert (approved):

CostinCarabas (approved):

PR #614, opened 2022-11-30
legacy proxy pair update user energy functionality

psorinionut:


sasurobert (approved):

locked-asset/proxy_dex/src/attr_ex_helper.rs
```diff
@@ -0,0 +1,76 @@
+use common_structs::LockedAssetTokenAttributes;
```
dorin-iancu, line 1: Make the module public in `factory` and import it from there, let's not copy-paste code unless absolutely necessary.

psorinionut: Done.

dorin-iancu (approved):

PR #617, opened 2022-12-02
properly update prop. total amount + code refactor

psorinionut:
- Moved the init function in the lib.rs file
- Properly update the proposal total amount when a user claims deposited tokens
- Updated the send function when a user claims deposited tokens, to only send the tokens once via direct_multi.


psorinionut:
Problems resolved in PR#618.

PR #627, opened 2022-12-05
farm with lock exit only

dorin-iancu:


dex/farm_with_lock/src/custom_rewards.rs
```diff
@@ -63,28 +61,4 @@ pub trait CustomRewardsModule:
 
         self.produce_rewards_enabled().set(&false);
     }
-
-    #[only_owner]
-    #[endpoint(setPerBlockRewardAmount)]
-    fn set_per_block_rewards(&self, per_block_amount: BigUint) {
```
sasurobert, line 69: can we stop the rewards if you delete this ? we want to set to 0.

sasurobert (approved):

PR #636, opened 2022-12-06
fix leftover locked tokens on exit farm with penalty

dorin-iancu:
- also remove compound rewards endpoint, since it's not possible anymore.

### DIFF
--- a/locked-asset/proxy_dex/src/energy_update.rs
+++ b/locked-asset/proxy_dex/src/energy_update.rs
@@ -1,5 +1,6 @@
 elrond_wasm::imports!();
 
+use common_structs::{Nonce, OldLockedTokenAttributes};
 use energy_factory::locked_token_transfer::ProxyTrait as _;
 use energy_query::Energy;
 use simple_lock::locked_token::LockedTokenAttributes;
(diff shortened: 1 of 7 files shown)


sasurobert (dismissed):

locked-asset/proxy_dex/src/proxy_farm.rs
```diff
@@ -208,6 +220,63 @@ pub trait ProxyFarmModule:
...
+            .contains(&proxy_farming_token.token_identifier);
+        if is_locked_token {
+            self.burn_locked_tokens_and_update_energy(
+                &proxy_farming_token.token_identifier,
+                proxy_farming_token.token_nonce,
+                &penalty_amount,
+                caller,
+            );
+
+            let mut remaining_locked_tokens = proxy_farming_token.clone();
```
psorinionut, line 251: This line is the same as the one from L257. Maybe compute it outside and use this variable, as it will be computed either way.

locked-asset/proxy_dex/src/energy_update.rs
```diff
@@ -8,25 +9,47 @@ static LOCKED_TOKEN_ID_STORAGE_KEY: &[u8] = b"lockedTokenId";
...
         let energy_factory_addr = self.energy_factory_address().get();
-        let locked_token_id = self.get_locked_token_id(&energy_factory_addr);
-        if token_id != &locked_token_id {
-            return;
+        let new_locked_token_id = self.get_locked_token_id(&energy_factory_addr);
+        if token_id == &new_locked_token_id {
+            let attributes: LockedTokenAttributes<Self::Api> =
+                self.get_token_attributes(token_id, token_nonce);
+            energy.update_after_unlock_any(token_amount, attributes.unlock_epoch, current_epoch);
+        } else {
```
psorinionut, line 44: For easier future updates, I would also put the check with legacy_locked_token_id here, otherwise it implies that it surely receives the legacy_locked_token, and I need to take a double check in the whole contract for that.

sasurobert (approved):

psorinionut (approved):

PR #637, opened 2022-12-06
fees collector grouped transfers

dorin-iancu:


energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -148,6 +149,41 @@ pub trait WeeklyRewardsSplittingModule:
...
+            first_payment.merge_with(second_payment);
+            merged_payments.push(first_payment);
+        }
+
+        merged_payments
+    }
+
+    fn clear_zero_payments(&self, payments: &mut PaymentsVec<Self::Api>) {
+        let mut i = 0;
+        let mut len = payments.len();
```
sasurobert, line 175: i think it is better to create a new payments vec and return that.

sasurobert: does it costs less gas like this ?

sasurobert (approved):

PR #639, opened 2022-12-07
Fees collector merging tokens

dorin-iancu:


sasurobert (commented):
I think it would be better with the old code, without grouping payments for the first launch. There is a lot of change in the last moments.

sasurobert:
No need for this as we send everything out through the multi transfer.

PR #648, opened 2022-12-08
cleanup proxy-dex token handling

dorin-iancu:


CostinCarabas (dismissed):

sasurobert (changes requested):
have to check if backward compatible.

PR #668, opened 2022-12-12
Return LKMEX from proxy-dex

CostinCarabas:
The extra MEX from impermanent loss will be locked for max period.

### DIFF
--- a/locked-asset/proxy_dex/src/lib.rs
+++ b/locked-asset/proxy_dex/src/lib.rs
@@ -36,6 +36,7 @@ pub trait ProxyDexImpl:
     + events::EventsModule
     + elrond_wasm_modules::default_issue_callbacks::DefaultIssueCallbacksModule
     + utils::UtilsModule
+    + locking_module::lock_with_energy_module::LockWithEnergyModule
 {
     /// asset_token_id: underlying asset token ID, which is used to interact with
     ///     pair/farm contracts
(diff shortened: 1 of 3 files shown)


locked-asset/proxy_dex/src/proxy_pair.rs
```diff
@@ -166,14 +168,19 @@ pub trait ProxyPairModule:
...
+
+            // Lock the difference from IL and send
+            let _ = self.lock_tokens_and_forward(
+                caller.clone(),
+                EgldOrEsdtTokenIdentifier::esdt(asset_token_id.clone()),
+                remaining_locked_amount,
+            );
 
             // burn base asset, as we only need to send the locked tokens
             self.send()
```
dorin-iancu, line 181: You should only burn `attributes.locked_tokens.amount - remaining_locked_amount`, since `remaining_locked_amount` is sent to the energy factory for locking.

locked-asset/proxy_dex/src/lib.rs
```diff
@@ -36,6 +36,8 @@ pub trait ProxyDexImpl:
     + events::EventsModule
     + elrond_wasm_modules::default_issue_callbacks::DefaultIssueCallbacksModule
     + utils::UtilsModule
+    + locking_module::locking_module::LockingModule
```
dorin-iancu, line 39: You should use `LockWithEnergyModule`, not `locking_module`, that's for simple-lock. `LockWithEnergyModule` is for energy-factory.

dorin-iancu: Add method for locking, similar to the one from `LockingModule`, currently, the `LockWithEnergyModule` is only used for virtual locking.

CostinCarabas:
continue in https://github.com/ElrondNetwork/sc-dex-rs/pull/671

PR #670, opened 2022-12-13
cleanup proxy-dex

dorin-iancu:
- make it clear that there are only 2 locked tokens. This makes further integrations easier.


sasurobert (dismissed):

energy-integration/common-modules/energy-query/src/lib.rs
```diff
@@ -38,7 +40,7 @@ pub trait EnergyQueryModule {
             return Energy::new_zero_energy(current_epoch);
         }
 
-        let energy_buffer: ManagedBuffer = self.read_storage_from_energy_factory(user);
+        let energy_buffer: ManagedBuffer = self.read_energy_from_factory(user);
```
psorinionut, line 43: Why not doing the whole check in the `read_energy_from_factory` function, so it just returns the energy? This way is way more easier to use in the future.

dorin-iancu: Everyone importing this module should use `get_energy_entry`. 

`read_energy_from_factory` is for internal use, and it's doing exactly what it says. Reading the energy. Validation can remain in this function imo, it brings no benefit moving it to the read function currently.

psorinionut (dismissed):

sasurobert (approved):

PR #676, opened 2022-12-14
Farm staking endpoints features

dorin-iancu:
- add the possibility of claiming boosted yields on enter farm-staking through additional payment
- add `exit_amount` argument for unstake endpoints
- add original_caller argument for proxy-endpoints, for proper boosted yields calculation
- move some code into the common farm-boosted-yields module


psorinionut (changes requested):

farm-staking/farm-staking-proxy/src/lib.rs
```diff
@@ -30,52 +30,31 @@ pub trait FarmStakingProxy:
...
-        require!(
-            lp_token_id.is_valid_esdt_identifier(),
-            "Invalide LP token ID"
-        );
+        self.require_valid_token_id(&staking_token_id);
+        self.require_valid_token_id(&lp_farm_token_id);
+        self.require_valid_token_id(&staking_farm_token_id);
+        self.require_valid_token_id(&lp_token_id);
 
         self.lp_farm_address().set(&lp_farm_address);
```
psorinionut, line 42: Why not `set_if_empty` here as well?

psorinionut (approved):

farm-staking/farm-staking-proxy/src/lib.rs
```diff
@@ -230,47 +203,29 @@ pub trait FarmStakingProxy:
...
-        if staking_rewards.amount > 0 {
-            user_payments.push(staking_rewards);
-        }
-        user_payments.push(unbond_staking_farm_token);
-
-        self.send().direct_multi(&caller, &user_payments);
-
-        user_payments.into()
+        let remaining_dual_yield_tokens =
+            EsdtTokenPayment::new(payment_token, payment_nonce, &payment_amount - &exit_amount);
```
psorinionut, line 218: There seems to be a wrong usage of parameters. Looks like exit_amount refers to the farm position, and it should not be deducted from the payment amount.

psorinionut (changes requested):

farm-staking/farm-staking-proxy/src/lib.rs
dorin-iancu: You're right, the current logic is wrong. Will merge with the fix once it enters the main branch.

farm-staking/farm-staking-proxy/src/lib.rs
dorin-iancu: Thought it might be a good idea to let those be overwritten, but it's probably best with set_if_empty here as well.

dorin-iancu: Done.

farm-staking/farm-staking-proxy/src/lib.rs
dorin-iancu: Fixed.

psorinionut (approved):

CostinCarabas (approved):

PR #683, opened 2022-12-19
old farms exit

dorin-iancu:
Don't merge this, or there will be consequences.


locked-asset/simple-lock/src/proxy_farm.rs
```diff
@@ -220,25 +239,26 @@ pub trait ProxyFarmModule:
...
-            lp_proxy_token_payment.token_nonce,
-            &lp_proxy_token_payment.amount,
+            initial_farming_tokens.amount,
         );
+        self.send()
+            .direct_non_zero_esdt_payment(&caller, &lp_proxy_token_payment);
+
+        let reward_tokens = exit_farm_result.get_reward_tokens();
+        self.send()
+            .direct_non_zero_esdt_payment(&caller, &reward_tokens);
```
sasurobert, line 249: It would be preferable to have one multiTransfer call.

locked-asset/simple-lock/src/proxy_farm.rs
```diff
@@ -177,6 +181,11 @@ pub trait ProxyFarmModule:
             &proxy_farm_token_attributes,
         );
 
+        self.send()
+            .direct_non_zero_esdt_payment(&caller, &enter_farm_result.reward_tokens);
```
sasurobert, line 185: One multiTransfer call would be preferable.

locked-asset/simple-lock/src/proxy_farm.rs
dorin-iancu: Done.

locked-asset/simple-lock/src/proxy_farm.rs
dorin-iancu: Most of the time, this is going to be a single payment, as rewards are more often than not 0 on enter. I'd rather keep it as mostly single payment and sometimes double payment, rather than add the overhead of multi-transfer for 1 payment, but it's your call.

sasurobert (approved):

CostinCarabas (approved):

sasurobert (approved):

PR #685, opened 2022-12-20
farm staking proxy exit amounts fix

dorin-iancu:


psorinionut (approved):

farm-staking/farm-staking-proxy/src/lib.rs
```diff
@@ -269,6 +294,10 @@ pub trait FarmStakingProxy:
         }
         user_payments.push(unbond_staking_farm_token);
 
+        if let Some(new_dual_yield_tokens) = opt_new_dual_yield_tokens {
+            user_payments.push(new_dual_yield_tokens)
+        }
+
         self.send().direct_multi(&caller, &user_payments);
 
         user_payments.into()
```
psorinionut, line 303: Wouldn't it be better if we let the payments vec as it is and use a `send_non_zero_payments` function? That way we can always return a reliable `user_payments` vector, which always contains the same number of payments (including the ones with 0 amounts). Right now, it may contain a different number of payments, for each unstake operation, and may be a bit harder to be further handled.

sasurobert (approved):

PR #690, opened 2022-12-22
add payments wrapper feature

psorinionut:


common/common_structs/src/wrapper_types.rs
```diff
@@ -92,3 +94,50 @@ impl<M: ManagedTypeApi> RawResultWrapper<M> {
...
+        Self { payments }
+    }
+
+    pub fn push(&mut self, payment: EsdtTokenPayment<M>) {
+        if payment.amount > 0 {
+            self.payments.push(payment);
+        }
+    }
+
+    pub fn send_to_address<T: SendApi>(&self, address: &ManagedAddress<M>) {
```
dorin-iancu, line 122: Remove `T` and just use `M`, it already has the `SendApi` bound.

common/common_structs/src/wrapper_types.rs
```diff
@@ -92,3 +94,50 @@ impl<M: ManagedTypeApi> RawResultWrapper<M> {
...
+                        &payment.token_identifier,
+                        payment.token_nonce,
+                        &payment.amount,
+                        0,
+                        &ManagedBuffer::new(),
+                        &ManagedArgBuffer::new(),
+                    );
+                }
+            }
+        }
```
dorin-iancu, line 137: Use multi-transfer here. Also, no need for `> 0` check, as you already do that when adding the payment.

common/common_structs/src/wrapper_types.rs
```diff
@@ -92,3 +94,50 @@ impl<M: ManagedTypeApi> RawResultWrapper<M> {
...
+    }
+
+    pub fn to_vec(&mut self) -> PaymentsVec<M> {
+        return self.payments.clone();
```
dorin-iancu, line 141:
```suggestion
    pub fn to_vec(self) -> PaymentsVec<M> {
        self.payments
```

common/common_structs/src/wrapper_types.rs
psorinionut: Updated.

common/common_structs/src/wrapper_types.rs
psorinionut: Updated.

common/common_structs/src/wrapper_types.rs
```diff
@@ -92,3 +94,44 @@ impl<M: ManagedTypeApi> RawResultWrapper<M> {
...
+                address,
+                &self.payments,
+                0,
+                &ManagedBuffer::new(),
+                &ManagedArgBuffer::new(),
+            );
+        }
+    }
+
+    pub fn into_vec(&mut self) -> PaymentsVec<M> {
```
dorin-iancu, line 134:
```suggestion
    pub fn into_vec(self) -> PaymentsVec<M> {
```

PR #693, opened 2022-12-28
add original caller functionality to simple lock

psorinionut:


CostinCarabas (dismissed):

sasurobert (approved):

locked-asset/simple-lock/src/farm_interactions.rs
```diff
@@ -62,10 +72,11 @@ pub trait FarmInteractionsModule {
         farming_token: TokenIdentifier,
         farming_token_amount: BigUint,
         additional_farm_tokens: ManagedVec<EsdtTokenPayment<Self::Api>>,
+        caller: &ManagedAddress,
     ) -> EnterFarmResultWrapper<Self::Api> {
         let mut contract_call = self
             .farm_proxy(farm_address)
-            .enter_farm()
+            .enter_farm(OptionalValue::Some(caller.clone()))
```
claudiulataretu, line 79: I don't think you need to clone the value

psorinionut: I've cloned the variable as the function receives a reference and the wrapping does not accept references. (I passed it as a reference to further use the variable in the main function).

psorinionut: Updated to clone the variable when calling the function.

CostinCarabas (approved):

sasurobert (approved):

PR #708, opened 2023-01-18
allow energy for SCs

psorinionut:
Allow SCs to have energy, following a set of rules:

- xMEX/LKMEX cannot be transferred to SCs through the ESCROW contract
- Wrapped XMEX cannot be unWrapped by SCs
- No unlock of LKMEX for SCs
- Migrate old tokens only adds energy to the contract, as it has no energy from where to deplete
- Contract can send out xMEX rewards only as wrapped tokens.


locked-asset/lkmex-transfer/src/lib.rs
```diff
@@ -46,6 +46,11 @@ pub trait LkmexTransfer:
     #[endpoint]
     fn withdraw(&self, sender: ManagedAddress) {
         let receiver = self.blockchain().get_caller();
+        require!(
+            !self.blockchain().is_smart_contract(&sender)
+                && !self.blockchain().is_smart_contract(&receiver),
+            "SCs cannot send or receive locked tokens"
+        );
```
dorin-iancu, line 53: Not needed here. If nobody can deposit from/to SCs, then they'd have nothing to withdraw anyway.

psorinionut: Removed.

dorin-iancu (approved):

CostinCarabas (approved):

PR #723, opened 2023-02-17
remove lkmex from locked token wrapper SC

psorinionut:
- Remove the possibility to send wrapped LKMEX.
- Small code refactor, removing the storage for the locked tokens and just read them from the energy factory SC.

### DIFF
--- a/locked-asset/locked-token-wrapper/src/lib.rs
+++ b/locked-asset/locked-token-wrapper/src/lib.rs
@@ -15,22 +15,8 @@ pub trait LockedTokenWrapper:
     + legacy_token_decode_module::LegacyTokenDecodeModule
 {
     #[init]
-    fn init(
-        &self,
-        old_locked_token_id: TokenIdentifier,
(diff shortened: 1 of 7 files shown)


alyn509 (approved):

locked-asset/locked-token-wrapper/src/lib.rs
```diff
@@ -71,17 +49,11 @@ pub trait LockedTokenWrapper:
         let payment = self.call_value().single_esdt();
         let original_locked_tokens = self.unwrap_locked_token(payment);
 
-        if original_locked_tokens.token_identifier == self.locked_token().get_token_id() {
+        if original_locked_tokens.token_identifier == self.get_locked_token_id() {
```
sasurobert, line 52: when is locked_token_id set ?

psorinionut: It is no longer saved. We read it directly from the energy factory storage.

PR #735, opened 2023-06-02
Fix merging locked tokens in case they have same unlock epoch

jacobleygonie:
Fixing an issue reported by the community, where merging 2 XMEX unlocking in the same month (e.g. March 2026) would give an XMEX unlocking at the next month (e.g. April 2026).
Also, adapting the weighted average calculation for the unlock epoch of an XMEX obtained when converting an LKMEX, to make sure the result is not underestimated.


sasurobert (approved):

sasurobert (approved):

psorinionut (approved):

ovidiuolteanu:
These change was taken over and incorporated into https://github.com/multiversx/mx-exchange-sc/pull/750 along with the full test suite for it.
No need to merge this specific PR anylonger.

PR #739, opened 2023-06-26
safe price v2 with distination address

psorinionut:


dex/pair/src/safe_price_view.rs
```diff
@@ -424,4 +505,24 @@ pub trait SafePriceViewModule:
 
         default_offset_rounds
     }
+
+    // legacy endpoints
```
CostinCarabas, line 509: Is there a reason for keeping legacy endpoints?

CostinCarabas (approved):

sasurobert (approved):

PR #745, opened 2023-07-10
safe price v2 README documentation

psorinionut:


dex/pair/README.md
```diff
@@ -213,3 +213,173 @@ The interaction scripts are located in the _dex/interaction_ directory. The scri
...
+
+The new safe price mechanism is an updated module of the __Pair SC__, designed to stabilize prices over a number of rounds, creating a smoother, more predictable pricing pattern in the __xExchange__ liquidity pools. The mechanism achieves this by storing accumulated reserves over time, with each recorded round representing a snapshot of token reserves at that point in time. This collection of reserve data is then processed through a variety of algorithms to compute and retrieve the safe price. Notably, the mechanism operates through a central view factory contract, which manages requests for all active smart contract pairs. The view factory contract accepts the address of the desired pair as a parameter in its view functions, simplifying the process of querying for data.
+
+__Important. The Safe Price module retrieves data independently of the liquidity pool's state. As such, even if a pair smart contract is paused, the safe price module will continue to return data. If an external contract requires more control or context concerning the data, it needs to first check the liquidity pool's status manually and subsequently utilize the safe price data according to its needs.__
```
ovidiuolteanu, line 224: Nice!

dex/pair/README.md
```diff
@@ -213,3 +213,173 @@ The interaction scripts are located in the _dex/interaction_ directory. The scri
...
+
+## Features
+
+- __Accumulated Reserves:__ The mechanism consistently stores the accumulated reserves over time. Each round creates a snapshot of the token reserves for that time, with the weight of a __PriceObservation__ for a round being the difference between the last saved round and the current one.
```
ovidiuolteanu, line 228:
```suggestion
- __Accumulated Reserves:__ The mechanism consistently stores the accumulated reserves over time. Each round in which the liquidity pool contract is used creates a snapshot of the token reserves for that time, with the weight of a __PriceObservation__ for a round being the difference between the last saved round and the current one.
```

dex/pair/README.md
```diff
@@ -213,3 +213,173 @@ The interaction scripts are located in the _dex/interaction_ directory. The scri
...
+- __Price Observations:__ Each round's reserves, once calculated and stored, are then saved in a __PriceObservation__ struct for the subsequent round (n+1). This allows for a clear record of price changes and liquidity over time.
+- __Circular List Storage and Binary Search:__ Price observations are stored in a circular list, which is an efficient data structure for storing the rolling price data. A binary search algorithm is used to find specific __PriceObservations__ in this list.
+- __Linear Interpolation:__ If a price observation was not saved for a requested round, the algorithm will perform a linear interpolation between the nearest price observations to estimate the price for that round.
```
ovidiuolteanu, line 231:
```suggestion
- __Linear Interpolation:__ If a price observation is not available for a queried round, the algorithm will perform a linear interpolation between the nearest price observations to estimate the price for that round.
```

dex/pair/README.md
```diff
@@ -213,3 +213,173 @@ The interaction scripts are located in the _dex/interaction_ directory. The scri
...
+- __Error Handling:__ To maintain data integrity, a request for a price observation older than the oldest stored observation will result in a SC error. This mechanism helps to prevent the use of outdated or non-existent data.
```
ovidiuolteanu, line 232:
```suggestion
- __Error Handling:__ To maintain data integrity, a query for a price observation older than the oldest stored observation will result in a SC error. This mechanism helps to prevent the use of outdated or non-existent data.
```

dex/pair/README.md
```diff
@@ -213,3 +213,173 @@ The interaction scripts are located in the _dex/interaction_ directory. The scri
...
+- __Versatile Safe Price Request Inputs:__ The mechanism offers several view functions, each providing a different way to request the safe price. These views give users flexibility in querying the safe price by either providing all necessary parameters or using default ones.
```
ovidiuolteanu, line 233:
```suggestion
- __Versatile Safe Price Request Inputs:__ The mechanism offers several view functions, each providing a different way to query the safe price. These views give users flexibility in querying the safe price by either providing all necessary parameters or using default ones. 
```

dex/pair/README.md
```diff
@@ -213,3 +213,173 @@ The interaction scripts are located in the _dex/interaction_ directory. The scri
...
+- __Circular List Storage and Binary Search:__ Price observations are stored in a circular list, which is an efficient data structure for storing the rolling price data. A binary search algorithm is used to find specific __PriceObservations__ in this list.
+- __Linear Interpolation:__ If a price observation was not saved for a requested round, the algorithm will perform a linear interpolation between the nearest price observations to estimate the price for that round.
+- __Error Handling:__ To maintain data integrity, a request for a price observation older than the oldest stored observation will result in a SC error. This mechanism helps to prevent the use of outdated or non-existent data.
+- __Versatile Safe Price Request Inputs:__ The mechanism offers several view functions, each providing a different way to request the safe price. These views give users flexibility in querying the safe price by either providing all necessary parameters or using default ones. 
+
+## Endpoints
```
ovidiuolteanu, line 235:
```suggestion
## Endpoints available on the View factory contract
```

dex/pair/README.md
```diff
@@ -213,3 +213,173 @@ The interaction scripts are located in the _dex/interaction_ directory. The scri
...
+    #[view(getSafePriceByTimestampOffset)]
+    fn get_safe_price_by_timestamp_offset(
+        &self,
+        pair_address: ManagedAddress,
+        timestamp_offset: u64,
+        input_payment: EsdtTokenPayment,
+    ) -> EsdtTokenPayment
+```
+
+A more specific view function, it allows you to retrieve the safe price by specifying the pair address, a timestamp offset, and input payment. It basically converts the timestamp_offset value to a generic round_offset, by dividing it with the constant value __SECONDS_PER_ROUND__.
```
ovidiuolteanu, line 294: Something to note for the future & timestamp related implementations: SECONDS_PER_ROUND is not a permanent constant. If protocol changes this (and I guarantee it will at some point), we'll start feeding incorrect averages until we upgrade the view contract.

dex/pair/README.md
```diff
@@ -213,3 +213,173 @@ The interaction scripts are located in the _dex/interaction_ directory. The scri
...
+        liquidity: BigUint,
+    ) -> MultiValue2<EsdtTokenPayment, EsdtTokenPayment>
+```
+
+This function allows you to simulate the value of both tokens within a liquidity pool based on a provided LP token amount. It receives the pair address, a timestamp_offset and the amount of LP tokens as parameters. Again, like the __getLpTokensSafePriceByRoundOffset__ endpoint, which automatically computes the start and end rounds of the query, this function calculates the round offset by dividing the timestamp_offset to a generic __SECONDS_PER_ROUND__ constant value.
+The function returns two output payments, one for each token in the pair, with their values computed at the safe price.
+
+## Legacy endpoints
+
+In order to avoid backwards compatibility issues, the two legacy endpoints from Safe Price V1 were kept, but they now use the new Safe Price V2 logic. One important aspect here is that they are not part of the Safe Price V2 view factory contract, but instead they are actual endpoints in the __Pair SC__.
```
ovidiuolteanu, line 359:
```suggestion
In order to avoid backwards compatibility issues, the two legacy endpoints from Safe Price V1 were kept, but they now use the new Safe Price V2 logic. One important aspect here is that they are not part of the Safe Price V2 view factory contract, but instead they are actual endpoints in the __Pair SC__.
__Important. These endpoints are planned to be phased out at any point in future, thus we consider them as deprecated and we recommend the usage of the ones available in the View factory contract.__
```

ovidiuolteanu (changes requested):

dex/pair/README.md
psorinionut: Added a information point:
__Important. The output of this endpoint (and any timestamp related endpoint) will return reliable data as long as the timestamp constant will remain unchanged at the protocol level.__

sasurobert (approved):

ovidiuolteanu (approved):

PR #761, opened 2023-08-16
Governance v2 energy updates

CostinCarabas:


energy-integration/governance-v2/src/views.rs
```diff
@@ -45,6 +45,19 @@ pub trait ViewsModule:
         }
     }
 
+    #[view(getUserVotingPower)]
+    fn get_user_voting_power(&self, user_address: ManagedAddress) -> BigUint {
+        let user_quorum = self.get_energy_amount_non_zero(&user_address);
+        user_quorum.sqrt()
```
dragos-rebegea, line 51: i would declare a method smoothing_function and use that everywhere. If the smoothing function will change there is only one place where the change will be applied

energy-integration/governance-v2/src/views.rs
```diff
@@ -45,6 +45,19 @@ pub trait ViewsModule:
         }
     }
 
+    #[view(getUserVotingPower)]
+    fn get_user_voting_power(&self, user_address: ManagedAddress) -> BigUint {
+        let user_quorum = self.get_energy_amount_non_zero(&user_address);
```
dragos-rebegea, line 50: I think it would be better no to have a required in the view, but return 0, if the user does not have energy.
Otherwise the vm query will return this:
```
{
	"data": {
		"data": {
			"returnData": null,
			"returnCode": "user error",
			"returnMessage": "No energy",
			"gasRemaining": 0,
			"gasRefund": 0,
			"outputAccounts": {},
			"deletedAccounts": null,
			"touchedAccounts": null,
			"logs": [
				{
					"identifier": "aW50ZXJuYWxWTUVycm9ycw==",
					"address": "erd1qqqqqqqqqqqqqpgqa7w2q8y9pma9k78kgtr3e5ndyzusj0h5czfs39xd2p",
					"topics": [
						"AAAAAAAAAAAFAO+coByFDvpbePZCxxzSbSC5CT70wJM=",
						"Z2V0VXNlclZvdGluZ1Bvd2Vy"
					],
					"data": "CglydW50aW1lLmdvOjExNzIgW2Vycm9yIHNpZ25hbGxlZCBieSBzbWFydGNvbnRyYWN0XSBbZ2V0VXNlclZvdGluZ1Bvd2VyXQoJcnVudGltZS5nbzoxMTcyIFtlcnJvciBzaWduYWxsZWQgYnkgc21hcnRjb250cmFjdF0gW2dldFVzZXJWb3RpbmdQb3dlcl0KCXJ1bnRpbWUuZ286MTE2OSBbTm8gZW5lcmd5XQ=="
				}
			]
		}
	},
	"error": "",
	"code": "successful"
}
```

CostinCarabas: Fixed in the following commit with `get_energy_amount` instead of `get_energy_amount_non_zero`.
For 0 cases, the smoothing function will handle it.

dragos-rebegea (approved):

PR #775, opened 2023-09-20
audit fixes (2)

psorinionut:


dex/farm/src/base_functions.rs
```diff
@@ -204,20 +206,68 @@ pub trait BaseFunctionsModule:
...
+
+        let attributes = FarmTokenAttributes {
+            reward_per_share: BigUint::zero(),
+            entering_epoch: 0,
+            compounded_reward: BigUint::zero(),
+            current_farm_amount: BigUint::zero(),
+            original_owner: self.blockchain().get_sc_address(),
+        };
+
+        let migration_farm_token_nonce = if farm_token_mapper.get_token_state().is_set() {
```
sasurobert, line 257: no need to create new token. Use VM endpoint GetCurrentESDTNFTNonce / implement it to Rust Framework if it does not exist.

farm-staking/farm-staking/src/claim_only_boosted_staking_rewards.rs
```diff
@@ -49,20 +49,37 @@ pub trait ClaimOnlyBoostedStakingRewardsModule:
...
             {
-                let mut user_total_farm_position = self.get_user_total_farm_position(caller);
-                user_total_farm_position.total_farm_position += farm_position.amount;
-                self.user_total_farm_position(caller)
-                    .set(user_total_farm_position);
+                migrated_amount += farm_position.amount;
             }
         }
+
+        if migrated_amount > 0 {
```
sasurobert, line 65: why do we have duplicated code ? Can't you add this into base function ?

farm-staking/farm-staking/src/lib.rs
```diff
@@ -144,4 +149,34 @@ pub trait FarmStaking:
...
+        let farm_token_mapper = self.farm_token();
+
+        let attributes = StakingFarmTokenAttributes {
+            reward_per_share: BigUint::zero(),
+            compounded_reward: BigUint::zero(),
+            current_farm_amount: BigUint::zero(),
+            original_owner: self.blockchain().get_sc_address(),
+        };
+
+        let migration_farm_token_nonce = if farm_token_mapper.get_token_state().is_set() {
```
sasurobert, line 167: Same as above - use GetCurrentESDTNFTNonce. Why do you have duplicated code ? Base function should resolv this - as code is exactly the same.

sasurobert (changes requested):

common/modules/farm/farm_base_impl/src/base_traits_impl.rs
```diff
@@ -196,12 +196,12 @@ pub trait FarmContract {
...
+            farm_token_mapper.require_same_token(&farm_position.token_identifier);
+
             if sc.is_old_farm_position(farm_position.token_nonce) {
                 continue;
             }
 
-            farm_token_mapper.require_same_token(&farm_position.token_identifier);
-
             let token_attributes: FarmTokenAttributes<<Self::FarmSc as ContractBase>::Api> =
                 farm_token_mapper.get_token_attributes(farm_position.token_nonce);
```
CostinCarabas, line 207: Can we optimize here the storage access for user_farm_position? After the for, settle with increase/decrease of user_farm_position.

dex/farm/src/base_functions.rs
psorinionut: Updated.

farm-staking/farm-staking/src/lib.rs
psorinionut: Updated.

farm-staking/farm-staking/src/claim_only_boosted_staking_rewards.rs
psorinionut: The farm staking contract does not import the base_functions module as there are some variables that are Farm only specific. Maybe we could rewrite this to be more generic.

common/modules/farm/farm_base_impl/src/base_traits_impl.rs
CostinCarabas, line 207: Can we extract the increase/decrease of the user_farm_position outside of the for, to improve the gas consumption?

dex/farm/src/base_functions.rs
```diff
@@ -204,20 +206,68 @@ pub trait BaseFunctionsModule:
...
-                user_total_farm_position.total_farm_position += farm_position.amount;
-                self.user_total_farm_position(caller)
-                    .set(user_total_farm_position);
+                migrated_amount += farm_position.amount;
             }
         }
+
+        if migrated_amount > 0 {
+            let mut user_total_farm_position = self.get_user_total_farm_position(caller);
+            user_total_farm_position.total_farm_position += &migrated_amount;
```
CostinCarabas, line 224: A gas improvement here could be an update instead of a get + set. (see the `decrease_old_farm_positions` function below).

`self.get_user_total_farm_position()` also verifies if position is empty, so a new function could do that (verification + update)

farm-staking/farm-staking/src/claim_only_boosted_staking_rewards.rs
CostinCarabas: We cannot do that due to Token Attributes. I think it's difficult to rewrite the code write now. The code changes will be huge and we would have to re-audit the whole solution.

farm-staking/farm-staking/src/lib.rs
CostinCarabas: The issue for the code duplication was `farm_token`. It has different attributes for farm and farm-staking.
I update the code: moved this to common module and provide the token_mapper as a parameter.

sasurobert (approved):

CostinCarabas (approved):

dex/farm/src/base_functions.rs
psorinionut: I don't know if this optimizes the gas cost as much, as we still need to check if the storage is empty before the update function. And also, this would mean another function, which would increase the contract size (even if by a small margin).

common/modules/farm/farm_base_impl/src/base_traits_impl.rs
psorinionut: I'll add it in the new PR.

psorinionut: Actually, it can be used only for increase position, as decrease position function needs the entire looped payment to check the attributes and decrease for the original owner. I'll see if there is any improvement if we send the original_owner as a parameter, instead of the payment.

PR #809, opened 2024-01-15
Fix: escrow transfers

CostinCarabas:
Fix a scenario where escrow transfers become unlockable at the time of withdrawal.


CostinCarabas:
Finish in https://github.com/multiversx/mx-exchange-sc/pull/810

PR #813, opened 2024-01-15
Escrow wrapper mainnet upgrade

CostinCarabas:


ovidiuolteanu:
We'll use this branch to release the applied fixes that came in with https://github.com/multiversx/mx-exchange-sc/pull/812
No need to merge this into the original release (let's keep the original release intact).

PR #818, opened 2024-01-25
hooks definition

dorin-iancu:
First draft of hooks.

### DIFF
--- a/dex/pair/src/pair_hooks/call_hook.rs
+++ b/dex/pair/src/pair_hooks/call_hook.rs
@@ -1,0 +1,47 @@
+use common_structs::PaymentsVec;
+
+use super::hook_type::{Hook, HookType};
+
+multiversx_sc::imports!();
+
+#[multiversx_sc::module]
+pub trait CallHookModule {
+    fn call_hook(
+        &self,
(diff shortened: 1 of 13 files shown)


dex/pair/src/pair_hooks/call_hook.rs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        let mut output_payments = input_payments;
+        for hook in &hooks {
+            let (_, back_transfers) =
+                ContractCallNoPayment::<_, MultiValueEncoded<ManagedBuffer>>::new(
+                    hook.dest_address,
+                    hook.endpoint_name,
+                )
+                .with_raw_arguments(call_args.clone())
```
CostinCarabas, line 35: I think the arguments should belong to `Hook` structure. At each call you are passing the same arguments.

dorin-iancu: Nope, I store the "Hook" in storage, so it needs to be as small as possible.

andreiblt1304 (approved):

mihaicalinluca (approved):

PR #835, opened 2024-02-12
allow setting first week epoch in farm

dorin-iancu:
Added the possibility of setting first_week_start_epoch when deploying the farm contracts.

Also done some misc changes:
- reorder some conditions for early-exit
- removed set_if_empty. No longer needed since we have a separate method
- also removed the migration code from the init function


CostinCarabas (approved):

dex/farm/src/base_functions.rs
```diff
@@ -304,15 +310,17 @@ where
...
-                let increase = (&split_rewards.base_farm * &storage_cache.division_safety_constant)
-                    / &storage_cache.farm_token_supply;
-                storage_cache.reward_per_share += &increase;
-            }
+        if total_reward == 0u64 {
+            return;
+        }
+
+        storage_cache.reward_reserve += &total_reward;
+        let split_rewards = sc.take_reward_slice(total_reward);
```
psorinionut, line 318: I would add a check in the `startProduceRewards` endpoint, to not allow the rewards to start generating, if current_epoch < first_week_start_epoch. This way, we avoid getting to the point where we may call an endpoint (like `enterFarm`) that calls the `take_reward_slice` function, which then would throw a SC error.

CostinCarabas (approved):

PR #843, opened 2024-02-21
Farm concentrated liq (part 3)

dorin-iancu:


dex/farm-concentrated-liq/src/cocentrated_liq.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub second_token_accumulated_fee: BigUint<M>,
+}
+
+pub struct PriceBounds<M: ManagedTypeApi> {
+    pub lower: BigUint<M>,
+    pub upper: BigUint<M>,
+}
+
+#[multiversx_sc::module]
+pub trait ConcentratedLiqModule:
```
psorinionut, line 29: Rename file to concentrated_liq.rs.

dex/pair/src/safe_price_view.rs
```diff
@@ -12,6 +13,12 @@ use crate::{
 pub const DEFAULT_SAFE_PRICE_ROUNDS_OFFSET: u64 = 10 * 60;
 pub const SECONDS_PER_ROUND: u64 = 6;
 
+#[derive(TypeAbi, TopEncode, TopDecode, NestedEncode, NestedDecode)]
+pub struct SafePriceLpToken<M: ManagedTypeApi> {
```
psorinionut, line 17: Why use a struct instead of MultiValue2?

dex/farm-concentrated-liq/src/cocentrated_liq.rs
```diff
@@ -0,0 +1,219 @@
...
+        if lower_diff < upper_diff {
+            price_bounds.lower
+        } else {
+            price_bounds.upper
+        }
+    }
+
+    fn get_price_bounds(&self, price: &BigUint) -> PriceBounds<Self::Api> {
+        let min_ticker = self.min_ticker().get();
+        let lower_bound = price / &min_ticker * &min_ticker;
```
psorinionut, line 147: I don't think this is correct

dex/farm-concentrated-liq/src/lib.rs
```diff
@@ -58,14 +61,18 @@ pub trait Farm:
...
     + utils::UtilsModule
+    + cocentrated_liq::ConcentratedLiqModule
 {
     #[init]
     fn init(
         &self,
         reward_token_id: TokenIdentifier,
         farming_token_id: TokenIdentifier,
         division_safety_constant: BigUint,
+        min_ticker: BigUint,
```
psorinionut, line 72: What is this? Does it need to be a variable? Maybe have it as a constant?

dex/pair/src/safe_price_view.rs
psorinionut: I would not change the implementation here, as there may be a lot of external contracts that have implemented this feature as it is right now.

dorin-iancu: Way easier to integrate into other contracts. You don't need to look into the source code to see what each multi-value means, you simply import the struct.

CostinCarabas (approved):

PR #850, opened 2024-03-04
Safe price with weighted LP supply

psorinionut:


dex/pair/src/safe_price_view.rs
```diff
@@ -466,14 +476,15 @@ pub trait SafePriceViewModule:
...
         }
     }
 
-    fn compute_weighted_reserves(
+    fn compute_weighted_amounts(
         &self,
         first_price_observation: &PriceObservation<Self::Api>,
         last_price_observation: &PriceObservation<Self::Api>,
-    ) -> (BigUint, BigUint) {
+    ) -> (BigUint, BigUint, BigUint) {
```
dorin-iancu, line 487: Really bad practice to return tuples, especially if all values have the same type. Use a struct instead.

psorinionut: Updated.

CostinCarabas (approved):

dorin-iancu (approved):

PR #852, opened 2024-03-08
accumulate boosted rewards

dorin-iancu:
Accumulating boosted rewards instead of claiming them when a user deposits, withdraws, claims base rewards, etc. Only case when boosted rewards should be claimed is when the user claims the boosted rewards (or they are auto-claimed for him).


psorinionut:
Not really sure this approach works as intended. Right now the flow is as follows:

- You accumulate boosted rewards in storage when the user reenters the farm or merges some tokens
- You can claim these boosted rewards (from enter and merge) only through `claimBoostedRewards` endpoint
- When claiming or exiting the farm, you compute the rewards like before, but without sending the collected rewards from enter and merge
- Basically you split the flow in 2, and my guess is that most of the time, this new storage will not be populated, as people usually claim the rewards through the claim endpoint

psorinionut:
I see two ways we can approach this:

1. We also distribute the accumulated rewards when distributing the base ones. Basically read the accumulated rewards from storage when calling the base claim endpoint, or exit farm
2. We completely split the compute rewards logic to return 2 different values `(base_farm_reward, boosted_yield_rewards)` instead of one single amount (`base_farm_reward + boosted_yield_rewards`). And we accumulate the `boosted_yield_rewards` in all endpoints (enter / merge / base claim / exit) and then simply send them when calling the dedicated endpoint. This would provide more code consistency.

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -122,12 +116,7 @@ pub trait Farm:
         &self,
         opt_orig_caller: OptionalValue<ManagedAddress>,
     ) -> ClaimRewardsResultType<Self::Api> {
```
psorinionut, line 118: Should we rename the endpoint to `claimBaseRewards`? To properly describe what happens inside the function? The same for the farm and farm-staking contracts.

dorin-iancu: Nah, people hate endpoint renames, especially the microservice guys.

energy-integration/farm-boosted-yields/src/lib.rs
```diff
@@ -142,9 +144,15 @@ pub trait FarmBoostedYieldsModule:
         }
     }
 
+    #[inline]
+    fn add_boosted_rewards(&self, user: &ManagedAddress, amount: &BigUint) {
```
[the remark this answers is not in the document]
dorin-iancu: Done.

CostinCarabas (approved):

PR #873, opened 2024-04-05
Router further audit improvements

psorinionut:


CostinCarabas (dismissed):

dex/router/src/config.rs
```diff
@@ -0,0 +1,69 @@
...
+        );
+    }
+
+    #[view(getPairCreationEnabled)]
+    #[storage_mapper("pair_creation_enabled")]
+    fn pair_creation_enabled(&self) -> SingleValueMapper<bool>;
+
+    #[view(getState)]
+    #[storage_mapper("state")]
+    fn state(&self) -> SingleValueMapper<bool>;
```
CostinCarabas, line 25: A better naming would make using this more easily.
For example, if `state = false` what does this mean?

dex/router/src/config.rs
```diff
@@ -0,0 +1,69 @@
+multiversx_sc::imports!();
+multiversx_sc::derive_imports!();
+
+use crate::{enable_swap_by_user::EnableSwapByUserConfig, factory::PairTokens};
+
+#[multiversx_sc::module]
+pub trait ConfigModule {
+    fn is_active(&self) -> bool {
+        self.state().get()
```
CostinCarabas, line 9: Couldn't you have use the pausable module?

dex/router/src/config.rs
psorinionut: Indeed, I agree that the naming isn't the most fortunate, but this is how it was before. I only moved the code in a different module, to be able to call some functions in the multi pair swap module.

dex/router/src/config.rs
psorinionut: The same as before, this was the code from the beginning and I would not change it because of backwards compatibility.

alyn509 (approved):

alyn509 (approved):

PR #886, opened 2024-04-10
Router check pair refactor

psorinionut:


CostinCarabas (dismissed):

dex/router/src/config.rs
```diff
@@ -2,18 +2,38 @@ multiversx_sc::imports!();
...
+        let first_token_id = self.get_first_token_id_mapper(pair_address.clone()).get();
+        let second_token_id = self.get_second_token_id_mapper(pair_address.clone()).get();
+
+        let pair_tokens = PairTokens {
+            first_token_id: first_token_id.clone(),
+            second_token_id: second_token_id.clone(),
+        };
+
+        let mut pair_map_address_opt = self.pair_map().get(&pair_tokens);
+        if pair_map_address_opt.is_none() {
```
claudiulataretu, line 23: you should not check for the reverse order of tokens. If the pair address is not found with the correct order provided by the pair storage, it will not work for sure

psorinionut: Removed.

psorinionut: Added back after feedback.

CostinCarabas (approved):

claudiulataretu (approved):

CostinCarabas (approved):

PR #938, opened 2024-08-01
farm supply for week fix

psorinionut:


dex/farm/src/lib.rs
```diff
@@ -228,10 +228,15 @@ pub trait Farm:
...
         }
 
+        let mut storage_cache = StorageCache::new(self);
+        Wrapper::<Self>::generate_aggregated_rewards(self, &mut storage_cache);
+
         let boosted_rewards = self.claim_only_boosted_payment(user);
         let boosted_rewards_payment =
             EsdtTokenPayment::new(self.reward_token_id().get(), 0, boosted_rewards);
 
+        self.set_farm_supply_for_current_week(&storage_cache.farm_token_supply);
```
sasurobert, line 238: do we have to do this every single time ?

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -237,7 +237,13 @@ pub trait Farm:
             );
         }
 
+        let mut storage_cache = StorageCache::new(self);
+        NoMintWrapper::<Self>::generate_aggregated_rewards(self, &mut storage_cache);
+
         let boosted_rewards = self.claim_only_boosted_payment(user);
+
+        self.set_farm_supply_for_current_week(&storage_cache.farm_token_supply);
```
sasurobert, line 245: do we have to do this every single time ?

dex/farm/src/lib.rs
psorinionut: We could also update all the claim endpoints to pass through an if statement.

psorinionut: We need to do this on claim, only when the storage is empty.

sasurobert: I think it would be better.

psorinionut: I thought about adding it but I don't think it will bring any real improvement. An if statement would require a few extra storage reads (we would need to compute the current week), so in the end gas won't be saved, but at the same time it would add some extra logic with the if statement, which in turn would complicate the review/audit/testing parts.

dex/farm-with-locked-rewards/src/lib.rs
psorinionut: Commented above. Also, this is how it already works for the other endpoints (enter, exit, claim).

sasurobert (approved):

PR #963, opened 2024-10-17
Move distribution to legacy

lcswillems:
System test:

https://github.com/lcswillems/xexchange-locked-asset-distribution-system-testing


psorinionut (approved):

legacy-contracts/locked-asset-distribution/src/lib.rs
```diff
@@ -42,15 +42,20 @@ pub trait Distribution {
     }
 
     #[endpoint(clearCommunityDistributionList)]
-    fn clear_community_distribution_list(&self, n: u64) -> usize {
-        for _ in 0..n {
-            self.community_distribution_list().pop_front();
+    fn clear_community_distribution_list(&self, n: u64) -> (u64, usize) {
+        let mut counter = 0;
```
psorinionut, line 46: Not really mandatory but would be good to have 2 extra things if we have polished the code.

1. rename `n` with something more suggestive, like `elements_to_delete`
2. Add a require that the passed variable (whatever the rename) is > 0. Right now, if you pass 0, it will still remove 1 element.

legacy-contracts/locked-asset-distribution/src/lib.rs
```diff
@@ -42,15 +42,20 @@ pub trait Distribution {
...
+                break;
+            }
         }
-        self.community_distribution_list().len()
+        (counter, self.community_distribution_list().len())
     }
 
     #[endpoint(clearUserLockedAssetMap)]
-    fn clear_user_locked_asset_map(&self, n: u64) -> usize {
+    fn clear_user_locked_asset_map(&self, n: u64) -> (u64, usize) {
```
psorinionut, line 58: Also, maybe add a comment about what is returned, like
// Returns number of processed elements and remaining elements in the storage.

psorinionut (changes requested):

psorinionut (dismissed):

psorinionut (approved):

psorinionut:
Addressed in PR #965

PR #966, opened 2024-10-22
Remove farm penalty code

dorin-iancu:


energy-integration/timestamp-oracle/src/epoch_to_timestamp.rs
```diff
@@ -0,0 +1,69 @@
...
+        self.timestamp_for_epoch(epoch).set(start_timestamp);
+    }
+
+    #[endpoint(updateAndGetTimestampStartEpoch)]
+    fn update_and_get_timestamp_start_epoch(&self) -> Timestamp {
+        let current_epoch = self.blockchain().get_block_epoch();
+        let last_update_epoch = self.epoch_last_interaction().get();
+        let mapper = self.timestamp_for_epoch(current_epoch);
+        if current_epoch == last_update_epoch {
+            return mapper.get();
```
psorinionut, line 31: Here we should always have a value. But because we have external factors, like the owner could update the storage with overwrite, I would also put a check here that the value stored by timestamp_for_epoch is not 0. And I would go even to a require, to completely crash the execution in that particular case, to not negatively impact the rewards computation.

energy-integration/farm-boosted-yields/src/custom_reward_logic.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        // Example: user entered at 25% of week, so give only 75% of rewards
+        let enter_timestamp_percent_of_week = linear_interpolation::<Self::Api, _>(
+            week_timestamps.start as u128,
+            week_timestamps.end as u128,
+            claim_progress.enter_timestamp as u128,
+            0,
+            MAX_PERCENT as u128,
+        );
+        let percent_leftover = MAX_PERCENT as u128 - enter_timestamp_percent_of_week;
```
psorinionut, line 157: Why all of this u128 converting?

energy-integration/farm-boosted-yields/src/custom_reward_logic.rs
```diff
@@ -0,0 +1,248 @@
...
+        core::cmp::min(max_rewards, boosted_reward_amount)
+    }
+
+    fn limit_boosted_rewards_by_claim_time(
+        &self,
+        user_reward: BigUint,
+        week_timestamps: &WeekTimestamps,
+        claim_progress: &ClaimProgress<Self::Api>,
+    ) -> BigUint {
+        if !(claim_progress.enter_timestamp >= week_timestamps.start
```
psorinionut, line 143: This is a bit hard to follow, especially negating the check. Could it be made a little bit easier to read?

psorinionut: The condition makes sense so we can keep it as it is, so we could have something like

```
claim_progress.enter_timestamp < week_timestamps.start
    || claim_progress.enter_timestamp >= week_timestamps.end
```

energy-integration/timestamp-oracle/src/epoch_to_timestamp.rs
psorinionut: I like the new implementation, where you don't need to terminate the execution, but return clearly that there is no data.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
```diff
@@ -165,3 +184,60 @@ pub trait WeeklyRewardsSplittingModule:
         }
     }
 }
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
dorin-iancu: It's actually common practice to have simple tests like this directly in code, but we haven't adopted it for some reason.

This test is very specific for the struct that's in this file, so I think it's okay for it to remain here.

energy-integration/farm-boosted-yields/src/custom_reward_logic.rs
dorin-iancu: Really? I would've thought it's actually easier to follow this way, since it's just "if not in this range, return".

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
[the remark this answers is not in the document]
psorinionut: I'm mixed feelings about this. It may be common practice, but this has nothing to do with the actual contract logic, and I much rather prefer to have as few lines as possible in the actual SC files. It's clear what's what at a first glance.

energy-integration/farm-boosted-yields/src/custom_reward_logic.rs
psorinionut: I don't think we could overflow. We have a maximum number of 6_048_000_000 (max number of seconds in a week * MAX_PERCENT), which is under u64 max limit.

energy-integration/farm-boosted-yields/src/custom_reward_logic.rs
psorinionut: We can keep it like this also, it's ok.

energy-integration/common-modules/weekly-rewards-splitting/src/lib.rs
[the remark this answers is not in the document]
dorin-iancu: I disagree actually. It's very clearly labeled as a test and it's in its own module too, so it's not polluting the implementation in any way. Maybe if it was at the beginning of the file I'd agree.

psorinionut (commented):
Looks good for me, but I think we should discuss how we organize the features branches. I'll green it after I take a look at all pending features and where we direct them.

psorinionut (approved):

psorinionut:
Should not be merged until we have a clear launch date.

PR #970, opened 2024-11-12
Disable add liquidity for locked tokens

dorin-iancu:


common/modules/disable-add-liq/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+    fn require_add_liq_enabled(&self) {
+        require!(
+            self.add_liq_disabled().get() == ADD_LIQ_ENABLED,
+            "Add Liquidity is disabled"
+        );
+    }
+
+    #[view(isAddLiqDisabled)]
+    #[storage_mapper("addLiqDisabled")]
+    fn add_liq_disabled(&self) -> SingleValueMapper<bool>;
```
CostinCarabas, line 31: I would rename this from `add_liq_disabled` to `add_liq_status` which will be set to `ADD_LIQ_DISABLED` or `ADD_LIQ_ENABLED`.

CostinCarabas (approved):

common/modules/disable-add-liq/src/lib.rs
dorin-iancu: Nah, then I'd have to specifically set it to `true` to have it enabled. I figured it's easier this way.

PR #971, opened 2024-11-12
change undistributed rewards code

dorin-iancu:


energy-integration/farm-boosted-yields/src/lib.rs
```diff
@@ -44,31 +54,41 @@ pub trait FarmBoostedYieldsModule:
...
-        self.require_caller_has_admin_permissions();
-
-        let collect_rewards_offset = USER_MAX_CLAIM_WEEKS + 1usize;
+    fn collect_undistributed_boosted_rewards(
+        &self,
+        start_week: Week,
+        end_week: Week,
+        dest_address: ManagedAddress,
+    ) {
+        let collect_rewards_offset = USER_MAX_CLAIM_WEEKS + 1;
```
CostinCarabas, line 65: Why use an offset? Why not claiming all the rewards from `start_week` to `end_week`?

locked-asset/energy-factory/src/unlocked_token_transfer.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    #[endpoint(transferUnlockedToken)]
+    fn transfer_unlocked_token(&self, dest: ManagedAddress, amount: BigUint) {
+        self.require_not_paused();
+        require!(amount != 0, "Invalid amount");
+
+        let caller = self.blockchain().get_caller();
+        require!(
+            self.unlocked_token_transfer_whitelist().contains(&caller),
+            "May not call this endpoint"
```
CostinCarabas, line 43: Maybe add a more descriptive error msg like Not authorized to call this endpoint.

CostinCarabas (approved):

energy-integration/farm-boosted-yields/src/lib.rs
dorin-iancu: It was like this in the initial implementation and I kept it. It makes sense: Users can claim boosted rewards for the past 4 weeks anytime, so you should only collect rewards from 5 weeks onward.

PR #972, opened 2024-11-13
Rc/v3.0.1.5

dorin-iancu:


common/modules/disable-add-liq/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+    fn require_add_liq_enabled(&self) {
+        require!(
+            self.add_liq_disabled().get() == ADD_LIQ_ENABLED,
+            "Add Liquidity is disabled"
+        );
+    }
+
+    #[view(isAddLiqDisabled)]
+    #[storage_mapper("addLiqDisabled")]
+    fn add_liq_disabled(&self) -> SingleValueMapper<bool>;
```
psorinionut, line 31: Even though we want to have it disabled, we shouldn't name the storage in a negative manner. I would suggest add_liq_enabled -> and set it to false. It is more clear this way. Also, this way the default would still be disabled, if the storage is empty.

locked-asset/proxy_dex/src/pair_interactions.rs
```diff
@@ -21,32 +39,33 @@ pub struct RemoveLiqudityResultWrapper<M: ManagedTypeApi> {
...
     fn call_add_liquidity(
         &self,
-        pair_address: ManagedAddress,
-        first_token_id: TokenIdentifier,
-        first_token_amount_desired: BigUint,
-        first_token_amount_min: BigUint,
-        second_token_id: TokenIdentifier,
-        second_token_amount_desired: BigUint,
-        second_token_amount_min: BigUint,
+        args: CallAddLiqArgs<Self::Api>,
```
psorinionut, line 42: I know this is way cleaner at a SC level, but isn't it more difficult for the one calling/integrating this endpoint? As he needs to also encode the struct, instead of just providing each element.

common/modules/disable-add-liq/src/lib.rs
dorin-iancu: Alright then. Currently working on something else, will do after.

locked-asset/proxy_dex/src/pair_interactions.rs
psorinionut: Indeed, you are right.

psorinionut (dismissed):
rc/v3.0.1 is not used any longer, it was merged to main. Anyway, this should not be merged until we have a clear launch date.

locked-asset/energy-factory/src/unlocked_token_transfer.rs
```diff
@@ -0,0 +1,55 @@
...
+        sc_addresses: MultiValueEncoded<ManagedAddress>,
+    ) {
+        let mapper = self.unlocked_token_transfer_whitelist();
+        for sc_addr in sc_addresses {
+            mapper.remove(&sc_addr);
+        }
+    }
+
+    #[endpoint(transferUnlockedToken)]
+    fn transfer_unlocked_token(&self, dest: ManagedAddress, amount: BigUint) {
```
psorinionut, line 36: The caller should not be given as a parameter. We should have an only_owner endpoint which sets the address in the energy factory storage. And we use only that. The farm should not have a saying in where the tokens are sent.

locked-asset/energy-factory/src/unlocked_token_transfer.rs
```diff
@@ -0,0 +1,55 @@
...
+        );
+
+        let base_asset_token_id = self.base_asset_token_id().get();
+        self.send()
+            .esdt_local_mint(&base_asset_token_id, 0, &amount);
+        self.send()
+            .direct_esdt(&dest, &base_asset_token_id, 0, &amount);
+    }
+
+    #[storage_mapper("ulkTokenTransfWhitelist")]
```
psorinionut, line 53: This storage is kinda like the same with the sc_whitelist_addresses storage, only that that storage has some extra contracts whitelisted in it. Not sure if we need the extra storage tho.

locked-asset/token-unstake/src/unbond_tokens.rs
```diff
@@ -20,41 +22,24 @@ pub trait UnbondTokensModule:
...
-                    let penalty_amount = &locked_tokens.amount - &unlocked_tokens.amount;
-                    if penalty_amount > 0 {
-                        let penalty = EsdtTokenPayment::new(
-                            locked_tokens.token_identifier,
-                            locked_tokens.token_nonce,
-                            penalty_amount,
-                        );
-                        penalty_tokens.push(penalty);
-                    }
+                    self.handle_single_unbond_entry(&entry, &mut penalty_tokens);
```
psorinionut, line 34: Keeping in mind that the previous version of token-unstake will go live in a few days, I don't think a new upgrade will be done any time soon. So I would avoid having any code refactor, as that will increase the difficulty in the release process, by bringing a new codehash. And anyways, it will not be deployed on mainnet.

common/modules/disable-add-liq/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+        self.add_liq_enabled().set(ADD_LIQ_ENABLED);
+    }
+
+    #[only_owner]
+    #[endpoint(disableAddLiq)]
+    fn disable_add_liq(&self) {
+        self.add_liq_enabled().set(ADD_LIQ_DISABLED);
+    }
+
+    fn require_add_liq_enabled(&self) {
```
psorinionut, line 22: What if this would act in a reverse way? Not add liquidity enabled, but add liquidity paused. So you have to write in storage whenever you want to pause the adding of liquidity. In other words, we would modify only this module, not the implementation in other contracts. We could keep require_add_liq_enabled as is, but check the storage has something instead of being empty.
This would be helpful when upgrading the contracts, so that the owner doesn't need to call the extra allow function.

locked-asset/energy-factory/src/unlocked_token_transfer.rs
dorin-iancu: Disagree. Not only farms will potentially use this, and they don't all have the same owner. We've discussed this previously.

locked-asset/energy-factory/src/unlocked_token_transfer.rs
dorin-iancu: Don't want to reuse whitelists. It makes the code very messy. Nobody knows what's there anymore, why it's there, etc.

locked-asset/token-unstake/src/unbond_tokens.rs
dorin-iancu: I really don't care about code hashes, and I refuse to keep bad code around just for that silly reason. There's literally nothing to test, it's the same code.

common/modules/disable-add-liq/src/lib.rs
dorin-iancu: ...I think it was you who told me to do it this way, and now you want it the other way again?

locked-asset/token-unstake/src/unbond_tokens.rs
```diff
@@ -64,10 +49,38 @@ pub trait UnbondTokensModule:
...
 
         self.send().direct_multi(&caller, &output_payments);
-
-        let new_unlocked_tokens = self.unlocked_tokens_for_user(&caller).get();
         self.emit_unlocked_tokens_event(&caller, new_unlocked_tokens);
 
         output_payments.into()
     }
+
+    fn handle_single_unbond_entry(
```
sasurobert, line 57: unstake refactor can enter. This is safe code.

energy-integration/energy-factory-mock/src/lib.rs
```diff
@@ -52,9 +52,23 @@ pub trait EnergyFactoryMock {
         self.user_energy(&user).set(&energy);
     }
 
+    #[endpoint(transferUnlockedToken)]
+    fn transfer_unlocked_token(&self, dest: ManagedAddress, amount: BigUint) {
```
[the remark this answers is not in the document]
dorin-iancu: This contract is just a mock, so it doesn't matter.

PR #990, opened 2024-12-11
[MEX-592] Distribute only mex in fees collector

dorin-iancu:


alyn509 (approved):

energy-integration/fees-collector/src/external_sc_interactions/pair.rs
```diff
@@ -0,0 +1,36 @@
+multiversx_sc::imports!();
+
+mod pair_proxy {
+    multiversx_sc::imports!();
+
+    #[multiversx_sc::proxy]
+    pub trait PairProxy {
+        #[payable("*")]
+        #[endpoint(swapTokensFixedInput)]
+        fn swap_tokens_fixed_input(
```
psorinionut, line 10: Why not integrate the actual Pair contract and write just a proxy? This way you can make a complete integration test.

energy-integration/fees-collector/src/fees_accumulation.rs
```diff
@@ -59,6 +55,37 @@ pub trait FeesAccumulationModule:
         }
     }
 
+    fn try_swap_to_base_token(&self, payment: &mut EsdtTokenPayment) {
```
psorinionut, line 58: I don't think this works as intended. We want to give all rewards in XMEX. This means we need to either go further by swapping the WEGLD to MEX and accumulate that. Or the other option would be, whenever we claim the rewards for a week, the first claim operation in the week does an extra swap of all the gathered WEGLD to MEX and then we distribuite the total MEX token.

psorinionut (changes requested):

energy-integration/fees-collector/src/external_sc_interactions/pair.rs
dorin-iancu: Dependency hell.

energy-integration/fees-collector/src/fees_accumulation.rs
dorin-iancu: Removed after latest specs.

energy-integration/fees-collector/src/lib.rs
```diff
@@ -29,32 +33,72 @@ pub trait FeesCollector:
...
+        let locked_token_id = self.get_locked_token_id();
+        let current_week = self.get_current_week();
+        for token_id in &all_tokens {
+            known_tokens_mapper.remove(&token_id);
+
+            if token_id == base_token_id || token_id == locked_token_id {
+                continue;
+            }
+
+            let acc_fees_mapper = self.accumulated_fees(current_week, &token_id);
```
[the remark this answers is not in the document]
psorinionut: I'd change this a bit, to make it completely backwards compatible. Instead of doing this, how about we take the entire balance of the SC and add them in the `all_accumulated_tokens` storage, for all tokens other than the base and locked tokens. We would also need to maybe clear the `accumulated_fees` storage for all the tokens, for the last 4 weeks before the upgrade.

psorinionut (approved):

energy-integration/fees-collector/tests/fees_collector_rust_test.rs
```diff
@@ -1594,20 +1442,212 @@ fn redistribute_rewards_test() {
...
+                        .get(),
+                    199
+                );
+            },
+        )
+        .assert_ok();
+}
+
+#[test]
+fn fees_collector_multiple_swap_test() {
```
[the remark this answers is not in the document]
psorinionut: Also, add a claim at the end, to check the swapped MEX tokens are distributed as intended and then check the users balance.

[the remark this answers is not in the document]
psorinionut: Add a new unit test where fees are gathered for a few weeks, then do the swap in week n, and then, still in week n, a new fee deposit of another token than the base one happens. We should check the rewards are distributed correctly, the storage are updated accordingly and the all_accumulated_tokens storage still holds the last deposit.

energy-integration/fees-collector/tests/fees_collector_rust_test.rs
```diff
@@ -1594,20 +1442,212 @@ fn redistribute_rewards_test() {
...
                     .get();
 
                 // i.e. 6 weeks worth of rewards minus what the third user claimed
                 assert_eq!(
                     first_token_balance,
                     managed_biguint!(599_952_417_140_485_515u64)
                 );
+            },
+        )
+        .assert_ok();
```
psorinionut, line 1455: Also check the storage is properly cleared for the past weeks, after the rewards redistribution.

psorinionut: Also, call `redistribute_rewards`  to see what happens.

psorinionut (changes requested):

energy-integration/fees-collector/src/lib.rs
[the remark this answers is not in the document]
dorin-iancu: Done.

energy-integration/fees-collector/tests/fees_collector_rust_test.rs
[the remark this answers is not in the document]
dorin-iancu: Only done first and third point, the second one takes way too much setup, and it's already proven through other tests.

energy-integration/fees-collector/tests/fees_collector_rust_test.rs
dorin-iancu: Done.

PR #997, opened 2025-02-04
Time block independent Safe Price

psorinionut:


dex/pair/src/safe_price.rs
```diff
@@ -47,10 +51,10 @@ impl<M: ManagedTypeApi> NestedDecode for PriceObservation<M> {
         let weight_accumulated = u64::dep_decode(input)?;
         let recording_round = u64::dep_decode(input)?;
 
-        let lp_supply_accumulated = if !input.is_depleted() {
-            BigUint::dep_decode(input)?
+        let (recording_timestamp, lp_supply_accumulated) = if !input.is_depleted() {
+            (u64::dep_decode(input)?, BigUint::dep_decode(input)?)
```
dorin-iancu, line 55: This will NOT work if you upgrade an existing pair with safe price. If you want to add new fields to a struct, always add them at the end, as they're encoded/decoded in that order.

dex/pair/src/safe_price.rs
```diff
@@ -100,7 +105,14 @@ pub trait SafePriceModule:
             new_index = (safe_price_current_index % MAX_OBSERVATIONS) + 1;
         }
 
-        if last_price_observation.recording_round == current_round {
+        let rounds_since_last_observation = current_round - last_price_observation.recording_round;
```
dorin-iancu, line 108: I'm not sure if this is correct for the very first observation, as that will always have round 0.

dex/pair/src/safe_price.rs
```diff
@@ -100,7 +105,14 @@ pub trait SafePriceModule:
...
+        let safe_price_round_save_interval_mapper = self.safe_price_round_save_interval();
+        let round_save_interval = match safe_price_round_save_interval_mapper.get() {
+            0 => DEFAULT_ROUND_SAVE_INTERVAL,
+            value => value,
+        };
```
dorin-iancu, line 113: You really don't need the mapper, you just care about the value. I'm also not a fan of this code style, avoiding if-else just for the sake of using the cool match.

```suggestion
        let mut round_save_interval = self.safe_price_round_save_interval().get();
        if round_save_interval == 0 {
                round_save_interval = DEFAULT_ROUND_SAVE_INTERVAL;
        }
```

This is way easier to understand imo. Also remember to format after applying suggestion, github editor sucks.

dorin-iancu: You could even go one step further and save `DEFAULT_ROUND_SAVE_INTERVAL` in the `safe_price_round_save_interval` in the init/upgrade functions. That way, you could get rid of this whole logic entirely.

dex/pair/src/safe_price.rs
```diff
@@ -143,14 +155,29 @@ pub trait SafePriceModule:
...
         new_price_observation.weight_accumulated += new_weight;
         new_price_observation.recording_round = new_round;
+        new_price_observation.recording_timestamp = self.blockchain().get_block_timestamp();
 
         new_price_observation
     }
 
+    #[only_owner]
+    #[endpoint(setSafePriceRoundSaveInterval)]
+    fn set_safe_price_round_save_interval(&self, new_interval: u64) {
```
dorin-iancu, line 165:
```suggestion
    fn set_safe_price_round_save_interval(&self, new_interval: Round) {
```

dex/pair/src/safe_price_view.rs
```diff
@@ -69,15 +69,33 @@ pub trait SafePriceViewModule:
...
+        let safe_price_current_index = self
+            .get_safe_price_current_index_mapper(pair_address.clone())
+            .get();
+        let price_observations = self.get_price_observation_mapper(pair_address.clone());
+
+        let target_observation = self.find_observation_by_timestamp(
+            target_timestamp,
+            safe_price_current_index,
+            &price_observations,
+        );
```
dorin-iancu, line 90: Duplicated code here

dex/pair/src/safe_price_view.rs
```diff
@@ -189,14 +207,32 @@ pub trait SafePriceViewModule:
...
```
dorin-iancu, line 227: And here

dex/pair/src/safe_price_view.rs
```diff
@@ -461,9 +497,14 @@ pub trait SafePriceViewModule:
         let lp_supply_sum = BigUint::from(left_weight) * left_observation.lp_supply_accumulated
             + BigUint::from(right_weight) * right_observation.lp_supply_accumulated;
 
+        let timestamp_sum = BigUint::from(left_weight) * left_observation.recording_timestamp
+            + BigUint::from(right_weight) * right_observation.recording_timestamp;
+        let interpolated_timestamp = (timestamp_sum / weight_sum).to_u64().unwrap_or_default();
+
```
dorin-iancu, line 503: Wasn't there a math module that did all the linear interpolation for you? This also applies to the lines above.

dex/pair/src/safe_price_view.rs
```diff
@@ -472,10 +513,76 @@ pub trait SafePriceViewModule:
...
+            recording_timestamp: interpolated_timestamp,
             lp_supply_accumulated,
         }
     }
 
+    fn find_observation_by_timestamp(
+        &self,
+        target_timestamp: u64,
+        current_index: usize,
+        price_observations: &VecMapper<Self::Api, PriceObservation<Self::Api>, ManagedAddress>,
```
dorin-iancu, line 525: I don't think you have to specify `Self::Api` here.

```suggestion
        price_observations: &VecMapper<PriceObservation<Self::Api>, ManagedAddress>,
```

dex/pair/src/safe_price_view.rs
```diff
@@ -472,10 +513,76 @@ pub trait SafePriceViewModule:
...
+                if target_timestamp > current_observation.recording_timestamp {
+                    target_timestamp - current_observation.recording_timestamp
+                } else {
+                    current_observation.recording_timestamp - target_timestamp
+                };
+
+            if current_timestamp_diff < min_timestamp_diff {
+                min_timestamp_diff = current_timestamp_diff;
+                closest_observation = current_observation.clone();
+            }
```
dorin-iancu, line 571: Can be its own function.

dex/pair/src/safe_price.rs
psorinionut: Good idea.

dex/pair/src/safe_price.rs
```diff
@@ -6,7 +6,9 @@ use multiversx_sc::codec::{NestedDecodeInput, TopDecodeInput};
 use crate::{amm, config, errors::ERROR_SAFE_PRICE_CURRENT_INDEX};
 
 pub type Round = u64;
+pub type Timestamp = u64;
 
+pub const DEFAULT_ROUND_SAVE_INTERVAL: u64 = 1;
```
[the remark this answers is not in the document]
psorinionut: Done.

dex/pair/src/safe_price.rs
psorinionut: Done.

dex/pair/src/safe_price_view.rs
psorinionut: Moved to another function.

dex/pair/src/safe_price_view.rs
psorinionut: Moved to another function.

dex/pair/src/safe_price_view.rs
psorinionut: Done.

dex/pair/src/safe_price_view.rs
dorin-iancu: Well, I personally think that making code easier to read is the most important thing :)

dex/pair/src/safe_price_view.rs
dorin-iancu: @psorinionut

dex/pair/src/lib.rs
```diff
@@ -99,10 +100,16 @@ pub trait Pair<ContractReader>:
             );
             self.add_permissions_for_all(admins, Permissions::ADMIN);
         };
+
+        self.safe_price_round_save_interval()
+            .set_if_empty(DEFAULT_ROUND_SAVE_INTERVAL);
```
dorin-iancu, line 105: It's always empty on init.

```suggestion
        self.safe_price_round_save_interval()
            .set(DEFAULT_ROUND_SAVE_INTERVAL);
```

dex/pair/src/safe_price_view.rs
psorinionut: Added weighted average from math module.

dex/pair/src/lib.rs
psorinionut: Done.

dex/pair/src/safe_price_view.rs
psorinionut: This doesn't work unless we pass around a lot of parameters, so I would avoid the extra complications. I've tried to do it as minimal as possible, but not succeeded.

dorin-iancu (approved):

ovidiuolteanu (changes requested):

dex/pair/src/safe_price_view.rs
```diff
@@ -2,16 +2,16 @@ multiversx_sc::imports!();
...
     amm, config,
     errors::{ERROR_SAFE_PRICE_OBSERVATION_DOES_NOT_EXIST, ERROR_SAFE_PRICE_SAME_ROUNDS},
     read_pair_storage,
-    safe_price::{self, PriceObservation, Round, MAX_OBSERVATIONS},
+    safe_price::{self, PriceObservation, Round, Timestamp, MAX_OBSERVATIONS},
 };
 
-pub const DEFAULT_SAFE_PRICE_ROUNDS_OFFSET: u64 = 10 * 60;
-pub const SECONDS_PER_ROUND: u64 = 6;
+pub const OFFSET_PRECISION_FACTOR: u64 = 1_000_000;
```
ovidiuolteanu: Is this used anywhere?

dex/pair/src/lib.rs
```diff
@@ -99,10 +100,22 @@ pub trait Pair<ContractReader>:
             );
             self.add_permissions_for_all(admins, Permissions::ADMIN);
         };
+
+        self.safe_price_round_save_interval()
+            .set(DEFAULT_ROUND_SAVE_INTERVAL);
     }
```
ovidiuolteanu, line 102: I think we're missing a set on default_safe_price_rounds_offset on init, since we don't use the hardcoded value anymore?

dex/pair/src/safe_price_view.rs
psorinionut: Removed.

dex/pair/src/lib.rs
psorinionut: Added the logic in router.

CostinCarabas (dismissed):

PR #1002, opened 2025-02-17
Max limit on unbond locked tokens

psorinionut:


andreiblt1304 (dismissed):

alyn509 (approved):

locked-asset/token-unstake/src/unbond_tokens.rs
```diff
@@ -2,6 +2,8 @@ use crate::events;
 
 multiversx_sc::imports!();
 
+pub const MAX_CLAIM_UNLOCKED_TOKENS: u64 = 30;
```
ovidiuolteanu, line 5: This limit is too high and it fails at protocol level as it exceeds the 100 builtin calls per tx limit.
Max limit in our case should be 20, consumes about 114,027,754 gas.

I think we have 5 builtin calls per unlocked token if burn > 0% and probably reduces to 4 in case of burn = 0%.

ovidiuolteanu (changes requested):

ovidiuolteanu (approved):

ovidiuolteanu (approved):

andreiblt1304 (approved):

PR #1006, opened 2025-02-19
Claim on behalf legacy tokens check

psorinionut:
Added a check to properly show to the users that they cannot use the claimOnBehalf endpoint using legacy tokens (that don't have the originalOwner value).

### DIFF
--- a/common/modules/original_owner_helper/src/lib.rs
+++ b/common/modules/original_owner_helper/src/lib.rs
@@ -23,6 +23,11 @@ pub trait OriginalOwnerHelperModule {
             let attributes: T = farm_token_mapper.get_token_attributes(payment.token_nonce);
             let payment_original_owner = attributes.get_original_owner();
 
+            require!(
+                !payment_original_owner.is_zero(),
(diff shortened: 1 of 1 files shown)


dorin-iancu (dismissed):

common/modules/original_owner_helper/src/lib.rs
```diff
@@ -23,6 +23,11 @@ pub trait OriginalOwnerHelperModule {
             let attributes: T = farm_token_mapper.get_token_attributes(payment.token_nonce);
             let payment_original_owner = attributes.get_original_owner();
 
+            require!(
+                !payment_original_owner.is_zero(),
+                "Cannot claim rewards with legacy positions"
```
ovidiuolteanu, line 28:
```suggestion
                "Cannot claim rewards on behalf of legacy positions"
```

ovidiuolteanu (changes requested):
let's change this one for clarity

andreiblt1304 (approved):

CostinCarabas (approved):

PR #1007, opened 2025-02-21
Token Unstake SC helper functions updates

psorinionut:
- Added setFeesBurnPercentage helper function
- Added a non zero burn penalty check


locked-asset/token-unstake/src/fees_handler.rs
```diff
@@ -67,13 +67,26 @@ pub trait FeesHandlerModule:
         self.burn_penalty(payment);
     }
 
+    #[only_owner]
+    #[endpoint(setFeesBurnPercentage)]
+    fn set_fees_burn_percentage(&self, fees_burn_percentage: u64) {
+        require!(
+            fees_burn_percentage <= MAX_PENALTY_PERCENTAGE,
+            "Invalid percentage"
```
andreiblt1304, line 75: I think this require message would be bit confusing for an user, maybe write something more verbose, such as:
"Fees burn percentage exceeds the maximum penalty percentage". 

Other than that, looks good to me!

andreiblt1304 (approved):

sasurobert (approved):

PR #1015, opened 2025-03-03
Owner claim undist farm boosted rewards

dorin-iancu:


energy-integration/farm-boosted-yields/src/lib.rs
```diff
@@ -30,6 +30,16 @@ impl<M: ManagedTypeApi> SplitReward<M> {
     }
 }
 
+mod energy_factory_proxy_send_rew {
```
psorinionut, line 33: Why not import the actual energy factory code? This way we can also do the integration tests. Si I would replace this with the actual reference.

energy-integration/farm-boosted-yields/src/lib.rs
```diff
@@ -44,31 +54,41 @@ pub trait FarmBoostedYieldsModule:
...
     #[endpoint(collectUndistributedBoostedRewards)]
-    fn collect_undistributed_boosted_rewards(&self) {
-        self.require_caller_has_admin_permissions();
-
-        let collect_rewards_offset = USER_MAX_CLAIM_WEEKS + 1usize;
+    fn collect_undistributed_boosted_rewards(
+        &self,
+        start_week: Week,
+        end_week: Week,
+        dest_address: ManagedAddress,
```
psorinionut, line 63: We don't need this, we will use the address saved in the energy factory (multisig).

locked-asset/energy-factory/src/unlocked_token_transfer.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        let base_asset_token_id = self.base_asset_token_id().get();
+        self.send()
+            .esdt_local_mint(&base_asset_token_id, 0, &amount);
+        self.send()
+            .direct_esdt(&dest, &base_asset_token_id, 0, &amount);
+    }
+
+    #[storage_mapper("ulkTokenTransfWhitelist")]
+    fn unlocked_token_transfer_whitelist(&self) -> WhitelistMapper<ManagedAddress>;
```
psorinionut, line 54: This should be an unordered set mapper, to be able to also iterate over this list. This will also be used later on by the governance SC.

psorinionut (approved):

PR #1021, opened 2025-03-11
Feat/claim farm undistributed rewards

psorinionut:


energy-integration/energy-factory-mock/src/lib.rs
```diff
@@ -52,9 +52,23 @@ pub trait EnergyFactoryMock {
         self.user_energy(&user).set(&energy);
     }
 
+    #[endpoint(transferUnlockedToken)]
```
CostinCarabas, line 55: There is no access control for this endpoint.
Is this expected behavior?

energy-integration/farm-boosted-yields/src/undistributed_rewards.rs
```diff
@@ -0,0 +1,102 @@
...
+            return total_rewards;
+        }
+
+        self.distribute_leftover_rewards(&total_rewards);
+
+        total_rewards
+    }
+
+    fn distribute_leftover_rewards(&self, total_rewards: &BigUint) {
+        let multisig_address = self.multisig_address().get();
```
CostinCarabas, line 72: check if storage `multisig_address` is set

psorinionut: THe implementation was changed in another PR, it was just merged. We no longer have this storage here in farms, we moved it in the energy-factory.

energy-integration/farm-boosted-yields/src/undistributed_rewards.rs
```diff
@@ -0,0 +1,73 @@
...
+        let base_token_id = self.get_base_token_id();
+        let reward_token_id = self.reward_token_id().get();
+        if base_token_id == reward_token_id {
+            let energy_factory = self.energy_factory_address().get();
+            let _: () = self
+                .energy_factory_proxy(energy_factory)
+                .transfer_unlocked_token(total_rewards.clone())
+                .execute_on_dest_context();
+        } else {
+            let owner = self.blockchain().get_caller();
```
CostinCarabas, line 61: You make an assumption here that the caller is the owner. Probably in the current setup, the endpoint is only owner. But this is a private function that may be called in another context.
The `owner` variable should be retrieved from `self.blockchain().get_owner_address()` instead.

energy-integration/energy-factory-mock/src/lib.rs
psorinionut: This SC is for test purposes only, as it is a mock contract. So there is no problem. Now, we've removed the mock contracts integration in tests, so it is no longer used. Not sure if we should keep it or remove it altogether.

sasurobert (approved):

andreiblt1304:
Is there a specific reason for the interactor github actions to be skipped for this PR ?

ovidiuolteanu:
@andreibancioiu removed them because something wasn't playing nice when running them. We'll have to investigate what should we do with them in a separate branch.

CostinCarabas (approved):

PR #1028, opened 2025-06-05
Fees collector block time independent logic

psorinionut:


energy-integration/fees-collector/src/lib.rs
```diff
@@ -78,5 +78,13 @@ pub trait FeesCollector:
             known_tokens_mapper.remove(&token_id);
             reward_tokens_mapper.insert(token_id);
         }
+
+        let locked_tokens_per_block_mapper =
+            SingleValueMapper::<Self::Api, BigUint>::new(StorageKey::new(b"lockedTokensPerBlock"));
+        let locked_tokens_per_block = locked_tokens_per_block_mapper.take();
+        let locked_tokens_per_epoch = locked_tokens_per_block * 10u64 * 60u64 * 24u64; // 14400 blocks per epoch
```
CostinCarabas, line 85: When/If the block time will change, these numbers will also change.
You should use some constant here declared in a config file or, ideally, in the framework. This way, further changes will be easier.

PR #1032, opened 2025-07-14
Timestamp based exchange SCs

psorinionut:
- Updated all SCs to comute/generate rewards based on timestamps instead of blocks. Covered SCs: Farm, Farm with locked rewards, Farm Staking, Router.
- Updated unit tests across the board.


common/modules/farm/farm_base_impl/src/base_traits_impl.rs
```diff
@@ -54,35 +54,35 @@ pub trait FarmContract {
         sc.send().esdt_local_mint(token_id, 0, amount);
     }
 
-    fn calculate_per_block_rewards(
+    fn calculate_per_second_rewards(
         sc: &Self::FarmSc,
-        current_block_nonce: Nonce,
-        last_reward_block_nonce: Nonce,
+        current_timestamp: Nonce,
+        last_reward_timestamp: Nonce,
```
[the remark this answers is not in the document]
psorinionut: Updated.

farm-staking/farm-staking/src/custom_rewards.rs
```diff
@@ -120,9 +120,9 @@ pub trait CustomRewardsModule:
         self.min_unbond_epochs().set(min_unbond_epochs);
     }
 
-    fn get_amount_apr_bounded(&self, amount: &BigUint) -> BigUint {
+    fn get_amount_apr_bounded(&self, amount: &BigUint, division_safety: &BigUint) -> BigUint {
```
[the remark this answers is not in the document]
psorinionut: Indeed, it affects the computations in the unit tests, but it doesn't mean it isn't correct. Furthermore, it is even more accurate to use the division safety constant, so that's why we chose this approach and updated the unit tests.

dex/farm-with-locked-rewards/src/lib.rs
```diff
@@ -88,12 +89,52 @@ pub trait Farm:
...
+        let last_reward_block_nonce_mapper =
+            SingleValueMapper::<Self::Api, u64>::new(StorageKey::new(b"last_reward_block_nonce"));
+        let per_block_reward_amount_mapper = SingleValueMapper::<Self::Api, BigUint>::new(
+            StorageKey::new(b"per_block_reward_amount"),
+        );
+
+        let per_block_reward_amount: BigUint<Self::Api> = per_block_reward_amount_mapper.take();
+        let last_reward_nonce = last_reward_block_nonce_mapper.take();
+
+        require!(
```
claudiulataretu: This should not be a require. For farms that would not produce rewards anymore the upgrade would fail. It should check for the produce rewards flag just to aggregate remaining rewards per block. Otherwise it should do nothing but still do the upgrade

dex/farm/src/lib.rs
```diff
@@ -90,12 +91,56 @@ pub trait Farm:
...
+        let last_reward_block_nonce_mapper =
+            SingleValueMapper::<Self::Api, u64>::new(StorageKey::new(b"last_reward_block_nonce"));
+        let per_block_reward_amount_mapper = SingleValueMapper::<Self::Api, BigUint>::new(
+            StorageKey::new(b"per_block_reward_amount"),
+        );
+
+        let per_block_reward_amount: BigUint<Self::Api> = per_block_reward_amount_mapper.take();
+        let last_reward_nonce = last_reward_block_nonce_mapper.take();
+
+        require!(
```
claudiulataretu: This should not be a require. For farms that would not produce rewards anymore the upgrade would fail. It should check for the produce rewards flag just to aggregate remaining rewards per block. Otherwise it should do nothing but still do the upgrade

farm-staking/farm-staking/src/lib.rs
```diff
@@ -96,19 +97,83 @@ pub trait FarmStaking:
         let current_epoch = self.blockchain().get_block_epoch();
         self.first_week_start_epoch().set_if_empty(current_epoch);
 
+        // Initialize last_reward_timestamp
```
claudiulataretu: This should be initialized only when the rewards start

farm-staking/farm-staking/src/lib.rs
```diff
@@ -96,19 +97,83 @@ pub trait FarmStaking:
...
+            SingleValueMapper::<Self::Api, u64>::new(StorageKey::new(b"last_reward_block_nonce"));
+        let per_block_reward_amount_mapper = SingleValueMapper::<Self::Api, BigUint>::new(
+            StorageKey::new(b"per_block_reward_amount"),
+        );
+
+        let current_block_nonce = self.blockchain().get_block_nonce();
+        let last_reward_nonce = last_reward_block_nonce_mapper.take();
+        let per_block_reward = per_block_reward_amount_mapper.take();
+
+        require!(
```
claudiulataretu: This should not be a require. For farms that would not produce rewards anymore the upgrade would fail. It should check for the produce rewards flag just to aggregate remaining rewards per block. Otherwise it should do nothing but still do the upgrade

farm-staking/farm-staking/src/lib.rs
```diff
@@ -96,19 +97,83 @@ pub trait FarmStaking:
...
+            let block_nonce_diff = current_block_nonce - last_reward_nonce;
+            &per_block_reward * block_nonce_diff
+        };
+
+        let farm_token_supply = self.farm_token_supply().get();
+        let max_apr = self.max_annual_percentage_rewards().get();
+        let extra_rewards_apr_bounded_per_block =
+            farm_token_supply * &max_apr / MAX_PERCENT / (SECONDS_IN_YEAR / 6u64);
+
+        let block_nonce_diff = current_block_nonce - last_reward_nonce;
```
claudiulataretu, line 126: You can move this line above the `extra_rewards_unbounded` and use inside the if statement

dex/farm-with-locked-rewards/src/lib.rs
psorinionut: Updated.

dex/farm/src/lib.rs
psorinionut: Updated.

farm-staking/farm-staking/src/lib.rs
psorinionut: Deleted.

farm-staking/farm-staking/src/lib.rs
psorinionut: Updated.

farm-staking/farm-staking/src/lib.rs
psorinionut: Updated.

claudiulataretu (approved):

PR #1041, opened 2025-09-09
safe price intermediate price functionality

psorinionut:


dex/pair/src/safe_price.rs
```diff
@@ -91,6 +91,72 @@ pub trait SafePriceModule:
...
+                lp_supply,
+            );
+            return;
+        }
+
+        // Check if enough rounds have passed since last finalized observation for direct save
+        let safe_price_current_index = self.safe_price_current_index().get();
+        let last_observation = if safe_price_current_index > 0 {
+            let price_observations = self.price_observations();
+            price_observations.get(safe_price_current_index)
```
dorin-iancu: Just use mapper directly, since you're not reusing it anyway.

```suggestion
            self.price_observations().get(safe_price_current_index)
```

dex/pair/src/safe_price.rs
```diff
@@ -127,6 +200,66 @@ pub trait SafePriceModule:
...
+            safe_price_current_index <= MAX_OBSERVATIONS,
+            ERROR_SAFE_PRICE_CURRENT_INDEX
+        );
+
+        let mut price_observations = self.price_observations();
+
+        let new_index = if price_observations.is_empty() {
+            1
+        } else {
+            (safe_price_current_index % MAX_OBSERVATIONS) + 1
```
dorin-iancu, line 249: Random question, but is this `safe_price_current_index` ever reset, or do we just naturally let it overflow and not care? I think you shouldn't "not care" unless `MAX_OBSERVATIONS` is a multiple of `usize::MAX` somehow.

dex/pair/src/safe_price_view.rs
```diff
@@ -517,69 +527,104 @@ pub trait SafePriceViewModule:
...
+            right_observation = price_observations.get(right_observation_index);
+        } else {
+            let left_observation_index = if search_index == 1 {
+                MAX_OBSERVATIONS
+            } else {
+                search_index - 1
+            };
+            left_observation = price_observations.get(left_observation_index);
+            right_observation = last_found_observation;
+        };
```
dorin-iancu: Not a fan of `let` bindings without assigning anything to them if it's not strictly needed by the code. Might need some formatting after though.

```suggestion
        let (left_observation, right_observation) = if last_found_observation.recording_timestamp < target_timestamp {
            let right_observation_index = (search_index % MAX_OBSERVATIONS) + 1;
            (last_found_observation, price_observations.get(right_observation_index))
        } else {
            let left_observation_index = if search_index == 1 {
                MAX_OBSERVATIONS
            } else {
                search_index - 1
            };
            (price_observations.get(left_observation_index), last_found_observation)
        };
```

dex/pair/src/safe_price.rs
psorinionut: Updated.

dex/pair/src/safe_price.rs
psorinionut: This resets. Whenever we get to the max value, it starts from 1 again, making it a round robin.

dex/pair/src/safe_price_view.rs
psorinionut: Updated.

dorin-iancu (approved):

PR #1042, opened 2025-09-12
Update ongoing proposal

CostinCarabas:


energy-integration/governance-v2/src/configurable.rs
```diff
@@ -68,6 +77,33 @@ pub trait ConfigurablePropertiesModule {
         self.try_change_voting_period_in_blocks(new_value);
     }
 
+    #[only_owner]
+    #[endpoint(changeProposalVotingPeriod)]
+    fn change_proposal_voting_period(&self, proposal_id: ProposalId, new_voting_period: u64) {
+        self.require_caller_not_self();
```
psorinionut, line 83: It's an only owner endpoint, i don't think require_caller_not_self applies.

energy-integration/governance-v2/src/configurable.rs
```diff
@@ -68,6 +77,33 @@ pub trait ConfigurablePropertiesModule {
...
+        self.require_valid_proposal_id(proposal_id);
+
+        // Ensure the proposal is currently Active
+        require!(
+            self.get_proposal_status(proposal_id) == GovernanceProposalStatus::Active,
```
psorinionut, line 88: Why not allow the Pending status as well?

alyn509 (approved):

PR #1057, opened 2025-10-08
Safe price timestamp search refactor

psorinionut:


dex/pair/src/lib.rs
```diff
@@ -109,6 +109,10 @@ pub trait Pair<ContractReader>:
     fn upgrade(&self) {
         self.safe_price_round_save_interval()
             .set_if_empty(DEFAULT_ROUND_SAVE_INTERVAL);
+
+        let default_safe_price_rounds_offset = 10 * 60;
```
CostinCarabas: Replace magic numbers "10", "60" with variables to improve code readability.

dex/pair/src/safe_price_view.rs
```diff
@@ -533,53 +532,112 @@ pub trait SafePriceViewModule:
...
+                right = mid - 1;
             }
         }
 
-        closest_observation
+        result_index
+    }
+
+    fn get_previous_observation_index(&self, current: usize, total_len: usize) -> usize {
+        if current > 1 {
```
CostinCarabas: I think you should go `current >= 1`

dex/pair/src/lib.rs
psorinionut: Updated.

dex/pair/src/safe_price_view.rs
psorinionut: Removed this code completely, changed the implementation.

Now, for this particular case, the code is correct to check for current > 1, as it refers to a vecmapper, where the index starts from 1. So current - 1 exists only for current = 2 or bigger.

